PR #3, opened 2017-03-13
Assignment week1

AbdulRahmanDbes:


rts-rob:
Hi Abdul Rahman,

Lots of good stuff in here! One question - from the assignment:
```Javascript
// Any other URL
// Response: return error code 404: Not found with a friendly message
```

What is the HTTP Response Code that you return if I hit (for example) `/invalid/url` ?

AbdulRahmanDbes:
Hi @rts-rob 
It should return an image, which I toke it from [Google](https://goo.gl/uu60ns)

`else print(response, request, "Error 404: Page Not Found");`

rts-rob:
Right... but what status code are you returning? 

See this link from the week 1 README for a hint: http://www.restapitutorial.com/httpstatuscodes.html

AbdulRahmanDbes:
@rts-rob Oh Right, I'm gonna fix it

Thanks ")

daanaerts:
Cool, looks good

PR #4, opened 2017-03-14
Node W1 HW Eyad

ekhamis:
Eyad HTTP

### DIFF
--- a/week1/index.js
+++ b/week1/index.js
@@ -1,60 +1,57 @@
-var http = require('http');
+const http = require('http');
 
 const hostname = '127.0.0.1';
 const port = 3000;
-var counter = 10;
-var requestLog = [];
+let counter = 10;
+let requestLog = [];
 
-const server = http.createServer(function(request, response) {
+const server = http.createServer((request, response) => {
     response.statusCode = 200;
     response.setHeader('Content-Type', 'text/html');
-    console.log('The new received http request is: ', request.url)
+    console.log('The new received http request is: ', request.url);
+
+    response.write(`<a href="/add"><button>Add</button></a> | `)
+    response.write(`<a href="/remove"><button>Remove</button></a> | `)
+    response.write(`<a href="/reset"><button>Reset</button></a> | `)
+    response.write(`<a href="/state"><button>State</button></a>`)
 
     switch (request.url) {
         case '/favicon.ico':
-            response.writeHead(200, { 'Content-Type': 'image/x-icon' });
-            console.log('favicon requested');
-            return;
+            response.end();
         case '/state':
         case '/':
-            console.log('Current counter value:', counter, '\n')
-            response.write('<h2>Counter current value is: ' + counter + '</h2>', 'utf8');
+            console.log(`Current counter value: ${counter} \n`)
+            response.write(`<h2>Counter current value is:  ${counter} </h2>`, 'utf8');
             requestLog.push(request.url)
             break;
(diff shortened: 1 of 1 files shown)


week1/index.js
```diff
@@ -1,25 +1,42 @@
...
+            counter = 10;
+            console.log('Counter value after RESET:', counter, '\n')
+            response.write(`<h2>Counter value was reseted to: ` + counter + `</h2>`, `utf8`);
+            break;
+        default:
+            response.statusCode = 404;
+            response.write(`<h2 style="color: red">Your request is invalid, please type either add, remove, state or reset to continue...<h2>`);
+            console.error('Error \n')
+    }
+    response.end();
```
rts-rob: Well done - exactly the right place for something that only executes once but _always_ executes.

week1/index.js
```diff
@@ -1,25 +1,42 @@
...
+            response.write(`<h2>The counter was successfully decreased by 1. <br> Counter new value is: ` + counter + `</h2>`, `utf8`);
+            break;
+        case '/reset':
+            counter = 10;
+            console.log('Counter value after RESET:', counter, '\n')
+            response.write(`<h2>Counter value was reseted to: ` + counter + `</h2>`, `utf8`);
+            break;
+        default:
+            response.statusCode = 404;
+            response.write(`<h2 style="color: red">Your request is invalid, please type either add, remove, state or reset to continue...<h2>`);
```
rts-rob: What's something else we could do here to make it easy for our users? They're in a browser, why type something when they could just click...

week1/index.js
```diff
@@ -1,25 +1,42 @@
...
-server.listen(port, function(error) {
-  if (error) {
-    console.log(error);
-  } else {
-    console.log('api listening on port', port);
-  }
+    switch (request.url) {
+        case '/state':
+        case '/':
+            console.log('Original counter value:', counter, '\n')
```
rts-rob: Try something:
1. Go to `/state`
2. Go to `/add`
3. Go back to `/state` (as a new request, not by hitting the back button

What gets logged to the console? Is that correct/true?

ekhamis: good catch. Corrected and will be pushed soon :)

week1/index.js
ekhamis: coming up :)

week1/index.js
ekhamis: :)

week1/index.js
```diff
@@ -1,25 +1,60 @@
-// To run this file: node index.js
-// Then in your browser: http://localhost:8080
 var http = require('http');
```
daanaerts: Hi Eyad,

- Great indentation, easy to read
- Question to think about: when to use `const`, `var`, or `let`?

week1/index.js
```diff
@@ -1,25 +1,60 @@
...
+    response.setHeader('Content-Type', 'text/html');
+    console.log('The new received http request is: ', request.url)
 
-// Start the HTTP server, start listening for requests
-server.listen(port, function(error) {
-  if (error) {
-    console.log(error);
-  } else {
-    console.log('api listening on port', port);
-  }
```
daanaerts: Interesting flow to handle request/response. I like it!

week1/index.js
ekhamis: Thanks noted. It was more of copy/paste blind spot, I have updated the code accordingly

PR #12, opened 2017-05-31
My homework 

massoodbarsa:
week 1
week 2

### DIFF
--- a/week2/index.js
+++ b/week2/index.js
@@ -4,7 +4,7 @@ var options = process.argv.slice(2);
 
 var command = options[0];
 
-
+//i have changed it
 switch (command) {
   case 'help':
     help();


rts-rob:
@massoodbarsa - Can you add the help file "help.txt" to this PR please?

week2/index.js
```diff
@@ -58,6 +101,78 @@ function listTodos() {
...
+function update() {
+  fs.readFile('todo.txt', 'utf8', function(err, data) {
+    if (err) {
+      throw err;
+    }
+    todo = splitStringByNewline(data)
+    let number = num()
+    todo[number - 1] = options[2]
+    let joinedLines = todo.join('\r\n').trim();
+    fs.writeFile('todo.txt', joinedLines);
```
rts-rob: You mean "fs.writeFileSync"? When I run your code I get:
`(node:1614) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.`

week2/index.js
```diff
@@ -58,6 +101,78 @@ function listTodos() {
...
+  })
+}
+
+//update a todo item with new text
+function update() {
+  fs.readFile('todo.txt', 'utf8', function(err, data) {
+    if (err) {
+      throw err;
+    }
+    todo = splitStringByNewline(data)
```
rts-rob: You haven't yet defined `todo` - best practice is to use `let todo = ...` here.

week2/index.js
```diff
@@ -4,14 +4,56 @@ var options = process.argv.slice(2);
...
+    remove()
+    break;
+  case 'reset':
+    reset()
+    break;
+  case 'update':
+    update()
+    break;
+  default:
+    showHelp();
```
rts-rob: Why not just `help()` as above?
Even better - how can we change this case to execute for both `help` and the default?
(DRY - don't repeat yourself)

rts-rob:
Good work here! I ran this through `eslint` and it raised a lot of warnings. Let's talk in class or via slack about how you can quickly format your code to remove those.

week2/index.js
```diff
@@ -4,14 +4,56 @@ var options = process.argv.slice(2);
...
+
+//find number after instruction
+function num() {
+  let number = parseInt(options[1])
+
+  //avoid type anything but numbers
+  if ((typeof number === 'number' && !isNaN(number))) {
+    return number
+
+  } else if (options[1] == null) {
```
jasongwartz: Careful with this `options[1] == null` comparison. While it does work right now, the actual value of an out-of-bounds array element is undefined, not null. So `options[1] === null` would not do what you want, whereas `options[1] === undefined` does, and is more likely to be what you intended for this line to mean.

PR #14, opened 2017-06-04
WEEK3 Home work

Odai-kakhi:
I have done with week3 homework

### DIFF
--- a/class7-week3/actions/clear.js
+++ b/class7-week3/actions/clear.js
@@ -1,0 +1,19 @@
+const Todo = require('../models/todo')
+
+module.exports = function clear (request, response) {
+
+
+    Todo.clear (error => {
+        if (error) {
+            console.error(error)
+            response.status(500)
+            response.json({
+                error: 'Internal error'
+            })
+        } else {
+            response.status(204)
+            response.end()
+        }
+    })
+
+}
--- a/class7-week3/actions/markAsDone.js
+++ b/class7-week3/actions/markAsDone.js
@@ -1,0 +1,22 @@
+const Todo = require('../models/todo')
+const deserializeTodo = require('../util/deserializeTodo')
+
+module.exports = function markAsDone(request, response) {
+    const id = request.params.id
+
+    Todo.markAsDone(id, (error, todo) => {
+        if (error) {
+            console.error(error)
+            response.status(500)
+            response.json({
+                error: 'Internal error'
+            })
+        } else {
+            response.status(201)
+            response.json({
+                todo
+            })
+        }
+    })
+
+}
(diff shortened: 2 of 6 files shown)


class7-week3/actions/clear.js
```diff
@@ -0,0 +1,19 @@
...
+
+    Todo.clear (error => {
+        if (error) {
+            console.error(error)
+            response.status(500)
+            response.json({
+                error: 'Internal error'
+            })
+        } else {
+            response.status(204)
```
rts-rob: nice

class7-week3/index.js
```diff
@@ -5,11 +5,23 @@ const app = Express()
...
+    clear,
+    markAsNotDone,
+    markAsDone
+} = require('./actions')
 
 app.get('/todos', list)
 app.post('/todos', create)
 app.put('/todos/:id', update)
 app.delete('/todos/:id', remove)
+app.delete('/todos', clean)
```
rts-rob: this should be `app.delete('/todos', clear)` right?

rts-rob: so you know, I can't run your code without making this change myself. you should never submit a PR that doesn't run. this results in "breaking the build" - something you don't want to do.

class7-week3/actions/clear.js
```diff
@@ -0,0 +1,19 @@
+const Todo = require('../models/todo')
+
+module.exports = function clear (request, response) {
+
+
+    Todo.clear (error => {
```
rts-rob: Conceptually, yes, but `clear is not a function`. If you want to do this you need to implement it in `models/todo` (a good idea)

class7-week3/actions/markAsNotDone.js
```diff
@@ -0,0 +1,22 @@
+const Todo = require('../models/todo')
+const deserializeTodo = require('../util/deserializeTodo')
+
+module.exports = function markAsNotDone(request, response) {
+    const id = request.params.id
+
+    Todo.markAsNotDone(id, (error, todo) => {
```
rts-rob: Again, good concept and return codes but this requires modifying the model.

class7-week3/actions/markAsDone.js
```diff
@@ -0,0 +1,22 @@
+const Todo = require('../models/todo')
+const deserializeTodo = require('../util/deserializeTodo')
+
+module.exports = function markAsDone(request, response) {
+    const id = request.params.id
+
+    Todo.markAsDone(id, (error, todo) => {
```
rts-rob: Again, good concept and return codes but this requires modifying the model.

PR #16, opened 2017-06-04
Homework week3

abuodh:
i have completed the week3 HomeWork

### DIFF
--- a/week2/index.js
+++ b/week2/index.js
@@ -40,7 +40,7 @@ function listTodos() {
       }
     }
 
-    var todos = splitStringByNewline(data);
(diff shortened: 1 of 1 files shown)


week2/index.js
```diff
@@ -39,8 +39,8 @@ function listTodos() {
         return console.log('Error: Something went wrong', error);
       }
     }
-
-    var todos = splitStringByNewline(data);
+ 
+    let todos = splitStringByNewline(data);
```
rts-rob: Good use of ES6 syntax!
Did you consider `const todos = ....` since it doesn't change?

PR #19, opened 2017-06-04
Add new function to todo function

AwsIdris:
Add new function to todo function


rts-rob:
I don't see any week3 homework in here; am I missing a PR?

PR #23, opened 2017-06-13
Completed assignment week 1

nos111:
Hi guys, Looking forward for your comments and reviews :)
Thanks alot


video.md
```diff
@@ -0,0 +1 @@
+hello nour
```
rts-rob: hello rob
:-)

week1/index.js
```diff
@@ -1,10 +1,71 @@
-// To run this file: node index.js
-// Then in your browser: http://localhost:8080
-var http = require('http');
+// Declare global variables
+const http = require('http');
```
jasongwartz: Good use of `const` and `let` here.

week1/index.js
```diff
@@ -1,10 +1,71 @@
...
+            state = 10;
+            operationsResponseHtml();
+            break;
+        default:
+            otherUrls();
+    }
+}
+
+//Function to increase the state by 1
+function addOne (){
```
jasongwartz: These functions take no parameters, yet modify global variables. Globals can be quite dangerous, especially in async programming.

What you probably want to do here is write functions which take arguments and return a modified output, so something more like:

```
function addOne (n) {
    return n++;
}
```
This is probably small enough functionality that you don't need a function for this, and could just use `++` inline.

week1/index.js
```diff
@@ -1,10 +1,71 @@
...
+    return state++;
+};
+//Function to decrease the state
+function subOne () {
+    return state--;
+};
+
+//Function to build the response HTML for subtraction and addition
```
jasongwartz: I love that you were verbose and clear. I did notice that your use of comments is effectively 'WHAT' the function does (which I can usually figure out from the function name), and not 'WHY' your functions do things in a certain way. It's worth spending a few extra lines explaining things clearly.

For example, in this case, I'd write something like:
```
// This function builds the HTML response for endpoints that change the value.
// Since the value doesn't need to be printed to acknowledge a change, this response
//    simply returns an acknowledgement without any content.
```

Good job on reading the assignment clearly, and just sending back the HTML with 'ok'. I didn't notice this until now, and you'll see when I share my example of the homework, I didn't follow the instructions clearly enough here 😬

week1/index.js
```diff
@@ -1,10 +1,71 @@
...
+// Declare global variables
+const http = require('http');
+//start server 
+const server = http.createServer();
+// the state which will be increased or decreased on the user request
+let state = 10; 
+//This port was already chosen by the assignment maker.
+const port = 8080;
+//Object containing the respons text and the content type.
+let responsToSendBack = {
```
jasongwartz: As I've commented elsewhere, the use of globals here is probably not the best approach. It's something new developers often do, including myself - on my first big programming project, I had to eventually go back and comb through a thousand lines of code to remove the dependencies on global variables.

See a very extensive description of why globals are bad here:
http://wiki.c2.com/?GlobalVariablesAreBad

What I would do instead is to pass the data back and forth through functions as arguments. This also ensures that the data will be consistent within a single request (what do you think will happen in your current code if two requests happen at the exact same time?).

PR #25, opened 2017-06-16
Implemented stateless functions and minimized global

nos111:
Hello guys,

I have tried to make my functions pass the data around and no more functions that takes no parameters and modifies global variables.

Also mimized my global variables use

Looking forward to hear your comments


week1/index.js
```diff
@@ -1,10 +1,86 @@
...
+    code:'<html></html>',
+    type:'text/html',
+    statusCode: 200,
+    state: Number
+};
 
-var port = 8080;
+//the original state which will always be 10. 
+//not allowed to change so it's a constant
+const originalState = 10;
```
jasongwartz: Great use of constant. Defaults are a perfect use-case for `const`.

A convention you'll sometimes see is the use of capital letters for constants, like `const ORIGINAL_STATE = 10`. This helps make a visual distinction between variable names and constant names.

week1/index.js
```diff
@@ -1,10 +1,86 @@
...
+function addOne (input){
+    return input += 1;
+};
+//Function to decrease the state by 1
+function subOne (input) {
+    return input -= 1;
+};
+
+//Function to build the response HTML for subtraction and addition
+//takes the whole response object and return the modified version.
```
jasongwartz: This is a great function 'docstring', which means a comment about what a function does. It explains what the function will do, what it takes in and what it returns out. You can also leave comments about why you implemented things in a certain way.

week1/index.js
```diff
@@ -16,10 +92,12 @@ server.listen(port, function(error) {
...
-  response.end();
-});
\ No newline at end of file
+    //send the URL to be checked for the prober response
+    //send the basic respons to the checkInput function to modify it
+    let inputToReturn = checkInput(request.url,responsToSendBack);
+    response.statusCode = inputToReturn.statusCode;
+    response.setHeader('content-type', inputToReturn.type);
+    response.write(inputToReturn.code);
+    response.end();
```
jasongwartz: This is a very well-designed listener handler - it's clear what the order of events are, and that all the content processing is done by another function. Great job. 😀

PR #26, opened 2017-06-16
Createed an http server that can add and subtract from state

yasershomaf:


week1/index.js
```diff
@@ -1,25 +1,62 @@
...
+            break;
+        case '/reset':
+            state = 10;
+            pText = 'The value of state was reset to 10.';
+            break;
+        default:
+            statusCode = 404;
+            h1Text = 'Oops!';
+            pText = "We can't seem to find the page you're looking for.";
+                       }
```
jasongwartz: The rest of your indenting and formatting was awesome up until this point 😛 Just be careful with this things, it can be tough on readability.

week1/index.js
```diff
@@ -1,25 +1,62 @@
...
-  } else {
-    console.log('api listening on port', port);
-  }
-});
+// This function is the handler for "request"
+// It understands the request, processes it
+// and then sends it to the handler for "response"
+function handleRequest(request, response) {
+    console.log('New http request received', request.url);
+    var statusCode = 200;// Set the default status code.
```
jasongwartz: Cool use of defaults, very sensible. You could also use `let` here, to make sure they don't accidentally get written in a different scope.

week1/index.js
```diff
@@ -1,25 +1,62 @@
-// To run this file: node index.js
-// Then in your browser: http://localhost:8080
+// Determine the port and the initial value of state
+var state = 10;
+var port = 8080;
```
jasongwartz: Does the port change during runtime? If not, this might be a good use of `const`.

week1/index.js
```diff
@@ -1,25 +1,62 @@
...
+// Start the HTTP server
 var http = require('http');
+var server = http.createServer(handleRequest);
 
-var port = 8080;
 
-var server = http.createServer();
+// This function is the handler for "response"
+// It sends the requested information to the user
+function handleResponse(response, statusCode, h1Text, pText) {
```
jasongwartz: Great use of a parameterised function here. This function does one job and only one job, which is very good design.

week1/index.js
```diff
@@ -1,25 +1,62 @@
...
+    response.writeHead(statusCode, {'content-type': 'text/html'});
+    response.write('<html><head></head><body><h1>' + h1Text + '</h1><p>' + pText + '</p></body></html>');
+    response.end();// End the response.
```
jasongwartz: I love your use of comments. This one might be a bit redundant, though. Comments should ideally help explain WHY you did things, not just what is happening.

week1/index.js
```diff
@@ -1,25 +1,62 @@
...
-    console.log('api listening on port', port);
-  }
-});
+// This function is the handler for "request"
+// It understands the request, processes it
+// and then sends it to the handler for "response"
+function handleRequest(request, response) {
+    console.log('New http request received', request.url);
+    var statusCode = 200;// Set the default status code.
+    var h1Text = 'OK';// Set the default text of the response title.
```
jasongwartz: Thank you for leaving a comment so I know that these are defaults! 😄

week1/index.js
```diff
@@ -1,25 +1,62 @@
...
-server.listen(port, function(error) {
-  if (error) {
-    console.log(error);
-  } else {
-    console.log('api listening on port', port);
-  }
-});
+// This function is the handler for "request"
+// It understands the request, processes it
+// and then sends it to the handler for "response"
```
jasongwartz: Great use of comments, thanks for helping me out to know what you're doing! :)

yasershomaf:
Home work for week 2:
Extract the argument and recognize it after making sure that it is a correct argument. If a wrong argument is entered, display a list of allowed arguments.

yasershomaf:
In case of using (help) as an argument or using empty argument, display the contents of help.txt

PR #28, opened 2017-06-17
pull request for my home work

gmamar:


week1/index.js
```diff
@@ -1,25 +1,61 @@
...
-  response.end();
-});
\ No newline at end of file
+// Create a event handler for "request" and calling render function. 
+
+server.on('request', function (request, response) {
+	response.setHeader('content-type', 'text/html');
+	switch (request.url) {
+		case '/':
+			info = 'Create HTML server';
```
jasongwartz: Your use of `info` as a global variable could be problematic (take a look at the comments I left for Nour in Slack and on his first pull request). What will happen if two requests come in at the exact same time? (Hint - this is called a race condition)

week1/index.js
```diff
@@ -1,25 +1,61 @@
...
+			info = 'Substract one from counter';
+			renderHtml();
+			break;
+		case '/reset':
+			state = 10;
+			info = 'Reset counter';
+			renderHtml();
+			break;
+		default:
+			info = 'Error code 404: Not found';
```
jasongwartz: The text says error code 404, but will this response actually send with an error code of 404? (Right now, I don't believe it will).

gmamar:
this is home work for week 2

week2/index.js
```diff
@@ -1,65 +1,129 @@
...
-switch (command) {
-  case 'help':
-  default:
-    showHelp();
-    break;
-  case 'list':
-    listTodos();
-    break;
+// find the value of the command, call help file in case command value not found.
+switch (options[0]) {
```
rts-rob: Just a style tip - when you assign the value of `options[0]` to a meaningful variable name like `command` it helps with clarity. 

So does your comment above, but in general in production code I like to avoid hard-coded things like `options[0]`. Again - just a question of style, not correctness.

week2/index.js
```diff
@@ -1,65 +1,129 @@
...
+	case 'help':
+	default:
```
rts-rob: Excellent nesting of related cases (lines 9 and 10)!

week2/index.js
```diff
@@ -1,65 +1,129 @@
...
+}
 
-    console.log('Your todo list looks like this');
-    todos.forEach(function(element, index) {
-      index = (index + 1).toString();
-      console.log(index, element);
-    });
+//delete one line from todo list by index number.
+function deleteLine() {
+	if (options[1] <= 0) { //Check if the line number is 0 and show help
```
rts-rob: What happens if the user has 4 items and enters `5` or greater as the line number? How could we check for that?

week2/index.js
```diff
@@ -1,65 +1,129 @@
...
+	options.shift(); //remove first element from command
+	var allLine = options.join(' ') ; //wrap all command as one line
+	fs.appendFileSync('todo.txt', allLine + '\n', 'utf8'); // append the new line to list
+	listTodos(); // show the new list
+}
 
-    var todos = splitStringByNewline(data);
+//update new todo line, creat a file in case its not exist.
+function updateLine() {
+	if (options[1] <= 0) { // in case that the user enter worng number
```
rts-rob: Same as line 95 - what if the user enters a line number that is too high? Or not a number/NaN?

week2/index.js
```diff
@@ -1,65 +1,129 @@
...
+	case 'reset':
+		resetFile();
+		break;
+	case 'update':
+		updateLine();
+		break;
 }
-
+//separate by new line file content, and filter it in array.
 function splitStringByNewline(string) {
```
rts-rob: Good use of what was built in - for an extra challenge, how could you convert this to read from/write to a JSON file?

week2/index.js
```diff
@@ -1,65 +1,129 @@
...
+}
 
-    var todos = splitStringByNewline(data);
+//update new todo line, creat a file in case its not exist.
+function updateLine() {
+	if (options[1] <= 0) { // in case that the user enter worng number
+		console.log('Please enter right line number!!! ' + '\n');
+		showHelp();
+	} else {
+		var lineNumber = options[1] - 1; // get the line number
```
jasongwartz: All of this relies on `options` being a global object. But what if we want `options` to be read/evaluated inside a function, and thus not in global scope - how would your code need to change?

week2/index.js
jasongwartz: Adding onto Rob's comment: remember that `options[1]` is a string, not an integer - you're actually relying on Javascript's implicit type coercion here. Try these lines out in a node interpreter to see what I mean:

`"-1" <= 0`

`"foo" <= 0`

week2/index.js
```diff
@@ -1,65 +1,129 @@
...
+		var data = fs.readFileSync('todo.txt', 'utf8') // get file information
```
jasongwartz: How many times in your code have you written `todo.txt`?

What happens if the filename where you save the todos needs to change? You're going to change it in all these places? What if it's written in 10,000 places?
How could we refactor your code to improve that?

PR #30, opened 2017-06-24
Home work week 2

yasershomaf:
Extract the argument and recognize it after making sure that it is a correct argument. If a wrong argument is entered, display a list of allowed arguments.
In case of using (help) as an argument or using empty argument, display the contents of help.txt

### DIFF
--- a/week2/help.txt
+++ b/week2/help.txt
@@ -1,25 +1,38 @@
-This is the help page for CLI ToDo app
-
 Instructions:
+=============
 
+
+help:
+-----
 To display this set of instructions:
 node index.js help
 
-To add an item to your todo's:
-e.g.
-node index.js add I need to brush my teeth
 
-To list all your todo's
-e.g.
+list:
+-----
+To list all your todo's:
 node index.js list
 
-To remove an item from your todo's 
-First list, and find the location (index, 0 based) of your todo
 
-To remove all items from your todo list
+add:
+----
+To add a new mission to your todo's:
+node index.js add "[your new mission]"
+
+
+remove:
+-------
+To remove an item from your todo's:
+node index.js remove [1 based index of the wanted item]
+
+
+reset:
+------
+To remove all items from your todo list:
 node index.js reset
 
-e.g.
-node index.js remove 0
 
-
+update:
+-------
+To update an existing item from your todo list with a new mission:
+node index.js update [1 based index of the wanted item] "[your new mission]"
--- a/week2/index.js
+++ b/week2/index.js
@@ -1,22 +1,29 @@
 // Extract the argument and recognize it after making sure that it is a correct argument. If a wrong argument is entered, display a list of allowed arguments. (done)
 // In case of using (help) as an argument or using empty argument, display the contents of help.txt (done)
-// In case of using (add) as an argument, make sure that the second argument is not empty then add it to todo.txt
+// In case of using (add) as an argument, make sure that the second argument is not empty then add it to todo.txt (done)
 // In case of using (remove) as an argument, make sure that the second argument is an acceptable index then show a warning message before removing the item from todo.txt
 // In case of using (update) as an argument, make sure that the second argument is an acceptable index and also the third argument is not empty then show a warning message before updating the item in todo.txt
 // In case of using (reset) as an argument, show a warning message before reseting todo.txt
 // In case of using (list) as an argument, show current todo's, or show an appropriate text if there are no todos.
 
 var fs = require("fs");
-var path = require("path");
 console.log("\n\n--<( ToDo App )>--\n\n")// Display a title with the app name.
 var arguments = process.argv.slice(2);// To extract the list of arguments
 switch (arguments[0]) {
     // In case of using "help" or empty argument
     case "help":
     case undefined:
-        console.log(fs.readFileSync(path.join(__dirname, "help.txt"), "utf8"));
+        console.log(fs.readFileSync(__dirname + "/help.txt", "utf8"));
         break;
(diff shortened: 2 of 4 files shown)


yasershomaf:
In case of using (add) as an argument, make sure that the second argument is not empty then add it to todo.txt

week2/help.txt
```diff
@@ -1,25 +1,38 @@
-This is the help page for CLI ToDo app
-
 Instructions:
```
rts-rob: Nice use of whitespace and clear instructions for the user!

week2/index.js
```diff
@@ -1,65 +1,46 @@
...
-    }
-  });
-}
-
-function openFile(fileName, callback) {
-  fs.readFile(__dirname + '/' + fileName, 'utf8', function(error, data) {
-    callback(error, data)
-  });
-}
+// Extract the argument and recognize it after making sure that it is a correct argument. If a wrong argument is entered, display a list of allowed arguments. (done)
```
rts-rob: Good stuff

week2/index.js
```diff
@@ -1,65 +1,46 @@
...
+// In case of using (help) as an argument or using empty argument, display the contents of help.txt (done)
+// In case of using (add) as an argument, make sure that the second argument is not empty then add it to todo.txt (done)
+// In case of using (remove) as an argument, make sure that the second argument is an acceptable index then show a warning message before removing the item from todo.txt
+// In case of using (update) as an argument, make sure that the second argument is an acceptable index and also the third argument is not empty then show a warning message before updating the item in todo.txt
+// In case of using (reset) as an argument, show a warning message before reseting todo.txt
+// In case of using (list) as an argument, show current todo's, or show an appropriate text if there are no todos.
+
+var fs = require("fs");
+console.log("\n\n--<( ToDo App )>--\n\n")// Display a title with the app name.
```
rts-rob: You can also do this with an ES6 string to see the formatting inline (but what you have is fine):
```javascript
console.log(`

--<( ToDo App )>--

`);
```

week2/index.js
```diff
@@ -1,65 +1,46 @@
...
+var arguments = process.argv.slice(2);// To extract the list of arguments
+switch (arguments[0]) {
+    // In case of using "help" or empty argument
+    case "help":
+    case undefined:
```
rts-rob: Good inclusion of "undefined" here. There's a way to move these down below the default case and only display the help file once. Can you figure it out?

week2/index.js
```diff
@@ -1,65 +1,46 @@
...
+// In case of using (add) as an argument, make sure that the second argument is not empty then add it to todo.txt (done)
+// In case of using (remove) as an argument, make sure that the second argument is an acceptable index then show a warning message before removing the item from todo.txt
+// In case of using (update) as an argument, make sure that the second argument is an acceptable index and also the third argument is not empty then show a warning message before updating the item in todo.txt
+// In case of using (reset) as an argument, show a warning message before reseting todo.txt
+// In case of using (list) as an argument, show current todo's, or show an appropriate text if there are no todos.
+
+var fs = require("fs");
+console.log("\n\n--<( ToDo App )>--\n\n")// Display a title with the app name.
+var arguments = process.argv.slice(2);// To extract the list of arguments
+switch (arguments[0]) {
```
rts-rob: Just a style tip - when you assign the value of `arguments[0]` to a meaningful variable name like `command` it helps with clarity.

A comment would also help, but in general in production code I like to avoid hard-coded things like `arguments[0]`. Again - just a question of style, not correctness.

week2/index.js
```diff
@@ -1,65 +1,46 @@
...
+        if (arguments[1]) {
+            fs.appendFileSync(__dirname + "/todo.txt", arguments[1] + "\n");
+            console.log("'" + arguments[1] + "' was added successfully to the todo's.");
+        }
+        else {
+            console.log("You have to write what you want to add to the todo's\ne.g., node " + __filename + " add 'Watch tv'");
+        }
+        break;
+    case "remove":
+        break;
```
rts-rob: So - looks like `remove`, `update`, `reset`, and `list` have not been completed.

Typically we would not create a PR until everything is ready to go.

I know that what happened here is that you're pushing to your own repository which already had an open PR. Pushing to the remote is a good practice and I don't want to discourage that.

The way to get around this is by using branches. We'll take some time in the next session to review git flow and GitHub and talk about feature branches.

week2/index.js
```diff
@@ -1,65 +1,46 @@
...
-    callback(error, data)
-  });
-}
+// Extract the argument and recognize it after making sure that it is a correct argument. If a wrong argument is entered, display a list of allowed arguments. (done)
+// In case of using (help) as an argument or using empty argument, display the contents of help.txt (done)
+// In case of using (add) as an argument, make sure that the second argument is not empty then add it to todo.txt (done)
+// In case of using (remove) as an argument, make sure that the second argument is an acceptable index then show a warning message before removing the item from todo.txt
+// In case of using (update) as an argument, make sure that the second argument is an acceptable index and also the third argument is not empty then show a warning message before updating the item in todo.txt
+// In case of using (reset) as an argument, show a warning message before reseting todo.txt
+// In case of using (list) as an argument, show current todo's, or show an appropriate text if there are no todos.
```
jasongwartz: Very glad to see that you wrote a spec here. Nice one!

week2/index.js
jasongwartz: +1 on this comment.

yasershomaf:
remove, update, reset, and list have been completed

PR #31, opened 2017-06-24
Essam New branch with new app  node.js w2

essamalsaloum:


week2/help.txt
```diff
@@ -1,25 +1,27 @@
-This is the help page for CLI ToDo app
 
-Instructions:
+This is the help page for CLI ToDo app
```
rts-rob: Nice full help file with examples! This is the kind of thing users like to see, and makes them feel better about your program.

week2/index.js
```diff
@@ -1,65 +1,131 @@
...
+var dataOfTodos = openFile('todo.txt');
+var todos = splitStringByNewline(dataOfTodos);
+//console.log(todos);
+console.log(`
+######||***** welcome *****||######
+######|| My List Todos App ||######
+`);
+//This function is to read the file and catch the error 
+//Which we used it above To assign the variable we will use it later
+function openFile(fileName) {
```
rts-rob: It's a really good practice to use helper functions like these.

Once you've written them, don't forget to use them! Try deleting the file `todo.txt` and then running `node index.js`. What happens? How can we fix this?

week2/index.js
```diff
@@ -1,65 +1,131 @@
...
-    });
-
-    if (todos.length > 5) {
-      console.log('You have too much to do!');
-    }
-  });
+//To add a new item.
+function addItem() {
+	//remove first item from process.argv
+	options.shift();
```
rts-rob: Very dangerous doing this on a global variable. What if the scope calling this function has a separate variable named "options" defined that is not what you expect?

What's another, safer way we can do this?

week2/index.js
```diff
@@ -1,65 +1,131 @@
...
 }
-
-function openFile(fileName, callback) {
-  fs.readFile(__dirname + '/' + fileName, 'utf8', function(error, data) {
-    callback(error, data)
-  });
+//To delete the selected item.
+function deleteLine() {
+	//Check for the item number
+	if ( valOfOptions > todos.length || valOfOptions <= 0) {
```
rts-rob: We always preach that whitespace is cheap, but if you're going to use the leading space _inside_ the parentheses like you've done here, you should add the trailing space as well: `if ( valOfOptions > todos.length || valOfOptions <= 0 )`

It's a question of style, and I prefer `if (valOfOptions > todos.length || valOfOptions <= 0)`, but the most important thing is to be consistent.

rts-rob:
Good job! I see some good thinking reflect in your code and your comments. Please keep those comments coming throughout your career as a developer; it makes it so much easier to understand what you're thinking.

week2/index.js
```diff
@@ -1,65 +1,131 @@
...
+######||***** welcome *****||######
+######|| My List Todos App ||######
+`);
+//This function is to read the file and catch the error 
+//Which we used it above To assign the variable we will use it later
+function openFile(fileName) {
+	try {
+		var fileContaint = fs.readFileSync(__dirname + '/' + fileName, 'utf8');
+	} catch (error) {
+		if (error.code === 'ENOENT') {
```
jasongwartz: What happens when I use this application for the very first time?

week2/index.js
```diff
@@ -1,65 +1,131 @@
...
+//To add a new item.
+function addItem() {
+	//remove first item from process.argv
+	options.shift(); 
+	//Make it on a single line
+	let newItem = options.join(' ') ;
+	// append the new item to list
+	fs.appendFileSync('todo.txt', '\n' + newItem + '\n', 'utf8');
+	//Open the file todo txt and read it again To update the information
+	dataOfTodos = openFile('todo.txt');
```
jasongwartz: At this point, we've opened and read the file, appended the file, and now we're opening and reading it again. What if the file is 1GB, or 100GB, or 100TB?

week2/index.js
jasongwartz: Good catch on the out of bounds - but have you accounted for off-by-one errors (because arrays are 0-based)?

week2/index.js
essamalsaloum: What is the main idea of this ?

week2/index.js
essamalsaloum: Is there a more useful way I need some hint ?

week2/index.js
jasongwartz: If I've never used your application before, the file won't exist. Will your application create it?

week2/index.js
jasongwartz: You've already got the contents of the file, and the new data.

If your intention is to validate that the write happened correctly, the strategy you're using is best.
If the point is just to show the newly-modified list to the user, you can print out the modified array instead of reading the file again. 😀

week2/index.js
essamalsaloum: I did not know it had to be created
Because it is originally inside the application folder
But that's true. thanks for the clarification

week2/index.js
essamalsaloum: A good idea, I will work on it

PR #34, opened 2017-07-07
Express

Ahalhowidi:


rts-rob:
Good use of branches!

class7-week3/actions/create.js
```diff
@@ -4,8 +4,8 @@ const deserializeTodo = require('../util/deserializeTodo')
 module.exports = function create(request, response) {
 
   const todo = deserializeTodo(request, response)
+  console.log(todo);
```
rts-rob: This type of debugging should come out before submitting a PR.

class7-week3/actions/clear.js
```diff
@@ -0,0 +1,16 @@
+const Todo = require('../models/todo')
+
+module.exports = function create(request, response) {
```
rts-rob: Good thinking to copy `create` - but you should rename it to `clear`. Also would have been easier to copy `remove`.

class7-week3/models/todo.js
```diff
@@ -75,6 +75,60 @@ class Todo {
     })
   }
 
+  clear(callback) {
+    this.load((error, todos) => {
```
rts-rob: This function is meant to remove all todos. Here you load the current todos, then you save them again in line 82. Does this accomplish what you want? (Hint: no, it doesn't)

class7-week3/actions/markAsDone.js
```diff
@@ -0,0 +1,23 @@
+const Todo = require('../models/todo')
+const deserializeTodo = require('../util/deserializeTodo')
+
+module.exports = function markAsDone(request, response) {
+
+  const id = request.params.id
+  const todo = deserializeTodo(request, response)
```
rts-rob: This got a little tricky to debug... Same comments apply to markAsNotDone.js

You should comment out lines 7 and 8. Those are only useful when you're extracting a todo from the body of a POST/PUT request. We aren't doing that here, so there's nothing to deserialize.

class7-week3/actions/markAsDone.js
```diff
@@ -0,0 +1,23 @@
...
+  if (todo == null) { return }
+
+  Todo.markAsDone(id, todo.done, (error, todo) => {
```
rts-rob: Once you've commented out lines 7 and 8, you'll need to look at this function signature. It's not correct.

Hint: Your functions in todo.js have the *correct* signature - so you should match them (again in markAsNotDone as well).

PR #38, opened 2017-09-02
node week1 hw

Ali-Barakat:


week1/resFun/addFn.js
```diff
@@ -0,0 +1,13 @@
+// import stateVar from './stateVar.js'
+// let state = 10
+// import state from './stateVar.js'
+
+export default function addFn(response){
+    response.setHeader('Content-Type', 'text/html')
+    response.write(`<html>
+    <title>HelloAliServer</title>
+    <body>
+    <div calss="main">Ok one has been added to state </div>
```
spirosikmd: You probably wanted to write `class` here!

week1/index.js
```diff
@@ -1 +1,57 @@
...
+        //     } else {
+                
+        //     }
+        }
+    
+    response.end()
+})
+server.listen(8080)
+
+console.log('sever is working')
```
spirosikmd: You could set your editor to automatically add [new line](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) at the end of the file on save!

week1/index.js
```diff
@@ -1 +1,57 @@
...
\ No newline at end of file
+import HTTP from 'http'
+import dflt from './resFun/dflt.js'
+import addFn from './resFun/addFn.js'
+import removeFn from './resFun/removeFn.js'
+import resetFn from './resFun/resetFn.js'
+// import Path from 'path'
+
+var state = 10
+let addOne = () => ++state
```
spirosikmd: I like these functions a lot! Very nice! You could have used `const addOne = () => ++state` instead of `let`, just to make sure the function value is never modified.

week1/index.js
```diff
@@ -1 +1,57 @@
...
+let reState = () => state = 10
+let subOne = () => --state
+
+const server = HTTP.createServer((request, response)=>{
+    console.log(request.method, request.url)
+    switch (request.url){
```
spirosikmd: A bit formatting here would help to make this function more readable!

week1/index.js
```diff
@@ -1 +1,57 @@
...
+        case '/':
+        dflt(response)
+        break
+        case '/state':
+        response.write(state.toString())
+        break
+        case '/add':
+        addOne()
+        addFn(response)
+        response.write(state.toString())
```
spirosikmd: Why do you call both `addFn` which writes the response, and `response.write` with `state` directly?

week1/index.js
```diff
@@ -1 +1,57 @@
...
+        case '/reset':
+        reState()
+        resetFn(response)
+        response.write(state.toString())        
+        break
+        
+        default:
+        response.statusCode = 404
+        response.setHeader('Contetn-Type', 'text/file')
+        response.write('404 error -- the file is not found- Fo')
```
spirosikmd: The user message could be a bit more helpful here. Users of this api, they are not actually requesting for a file. They are performing an action. And you could still return as `text/html`!

PR #39, opened 2017-09-03
Homework week1

AlexanderSaid:


week1/index.js
```diff
@@ -1,38 +1,81 @@
+
+var state = 10
+var err = "This page does not exist, please insert correct path name"
+
 import HTTP from 'http'
-import Path from 'path'
+// import stateHtml from './stateHtml'
+// I have tried to put this function in another file 
+// but it keeps tell me that a [state] is undifined
```
spirosikmd: Putting this function in another file is a good idea. You get this error because if you just move the function as is now in another file, `state` is indeed undefined. However, if you pass a state variable to the function, for example

```javascript
function stateHtml(response, state) { /** function body here */ }
```

then you can put it in another file, import it and use it as `stateHtml(response, state)`. Does this make sense?

week1/errorMessage.js
```diff
@@ -0,0 +1,4 @@
+export default function errorMessage(response, err) {
+    response.setHeader('Content-Type', 'text/plain');
+    response.write(err);
+}
```
spirosikmd: You could set your editor to automatically add [new line](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) at the end of the file on save!

PR #40, opened 2017-09-03
ayman hw

aymansalah1:


week1/index.js
```diff
@@ -1 +1,37 @@
-console.log("Hello world")
\ No newline at end of file
+import Path from 'path'
+import HTTP from 'http' 
+
+import sendIndexHTML from './responses/sendIndexHTML'
+import sendText from './responses/sendText'
+var state = 10 
+
+const server = HTTP.createServer ((request,response)=>{
```
spirosikmd: Formatting in this function could be a bit better to make it more readable!

week1/responses/sendIndexHTML.js
```diff
@@ -0,0 +1,6 @@
+export default function sendIndexHTML (response,state,calculation){
```
spirosikmd: You could simplify this function to only send the state when a state is passed. For example:

```javascript
export default function sendIndexHTML (response, state) {
  response.setHeader('Content-type','text/html')
  response.write(
    `<html><header><title>calculator</title></header><body>${state || 'ok'}</body></html>`
  )
}
```

aymansalah1:
ayman's HW week 2 nodeJS

PR #42, opened 2017-09-06
somur's Home work

ruteeb:


ruteeb:
k

week1/pages/add.js
```diff
@@ -0,0 +1,16 @@
...
+ 
+ 
+
+module.exports = {
+    "body" : function contentHomePage(Response ){
+        STATUS.varia = STATUS.varia + 1 ;
+        Response.write('<h1>add one</h1>')
+        console.log('this is the value of varia '+ STATUS.varia)
+        Response.write(STATUS.varia.toString())
+        Response.end();
```
spirosikmd: You are calling `response.end()` in every page, but also in the `index.js`. Maybe you could remove it from the page functions. It should still work, right?

week1/pages/add.js
```diff
@@ -0,0 +1,16 @@
+
+import HTTP from 'http'
+import STATUS  from './status.js'
+ 
+ 
+
+module.exports = {
+    "body" : function contentHomePage(Response ){
```
spirosikmd: I think you complicated yourself a bit with this design. You could just export `function contentHomePage(Response) { ... }` and it should still work right?

week1/pages/add.js
```diff
@@ -0,0 +1,16 @@
...
+        STATUS.varia = STATUS.varia + 1 ;
```
spirosikmd: I think it would be more maintainable if these page function didn't modify the state variable. You could pass the state variable as an extra argument. For example,

```javascript
function contentHomePage(response, state) {
  /** function body here */
}
```

What do you think? Does it make sense?

ruteeb:
Yes thank you very much

PR #43, opened 2017-09-08
homework week 2 added

2sma3il1991juve:


spirosikmd:
Hi @2sma3il1991juve! Would be great if you ignore the node modules from this pull request, so it is easier to review! Thanks a lot!

PR #44, opened 2017-09-09
homeworks added without node modules 

2sma3il1991juve:


week1/index.js
```diff
@@ -1 +1,34 @@
...
+            resetHTML(response, state)
+            break
+        default:
+            response.statusCode = 404
+            notFound(response, "File not found")
+            break
+    }
+    response.end()
+})
+server.listen(8080)
```
spirosikmd: You could set your editor to automatically add [new line](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) at the end of the file on save!

week1/responses/notFound.js
```diff
@@ -0,0 +1,4 @@
+export default function notFound(response, text) {
+	response.setHeader('Content-Type', 'text/plain')
+	response.write(text)
```
spirosikmd: The *not found* response could still be `text/html`!

week2/help.txt
```diff
@@ -0,0 +1,5 @@
+Welcome to the help guide, please use one of the following commands
+list: to show the todo-list
+add: to add to the todo-list
+remove: to remove an an item from the todo-list
```
spirosikmd: Extra `an` here 😄

week2/index.js
```diff
@@ -0,0 +1,41 @@
+import fs from 'fs';
+const argumentArray = process.argv;
+const command = process.argv[2];
+const argument = process.argv[3];
+switch (command) {
+  case "help":
+  case undefined:
```
spirosikmd: Instead of checking for `case undefined`, you can execute this on `default:`. For example:

```javascript
switch(command) {
  // other commands here
  case 'help':
  default:
    // show help
}
```

Would this work?

week2/index.js
```diff
@@ -0,0 +1,41 @@
...
+    fs.readFile('help.txt', 'utf8', (err, data) => {
+      if (err) throw err;
+      console.log(data);
+    });
+    break;
+  case "add":
+    fs.appendFile('todoList.txt', "\n" + argument, (err) => {
```
spirosikmd: The problem with appending `\n` before `argument` is that the result of this command `node index.js add "task 5"` would be 

```
task 0
task 1
task 2
task 3
task 4

task 5
```

the first time we use the app. So we don't really want to add a newline character before, but after the `argument`. To make sure that it is always there!

week2/index.js
```diff
@@ -0,0 +1,41 @@
...
+      if (err) throw err;
+      console.log('The task is added successfully');
+    });
+    break;
+    case "remove":
+    fs.readFile('todoList.txt', 'utf8', (err, data) => {
+      if (err) throw err;
+      const dataToArray = data.split("\n");
+      dataToArray.splice(argument, 1);
```
spirosikmd: Homework specifies that we should `remove a todo item by its 1-base index`. Here you are using the 0-index. Which means that if we execute `node index.js remove 1` the second todo item will be removed, while we wanted the first todo item to be removed. How can we fix that?

PR #50, opened 2017-10-29
The HomeWork for this week

itemah26:


week1/HomeWork/index.js
```diff
@@ -0,0 +1,33 @@
...
+          state += 1;
+          response.write(`<h1>You requested the path: ${ request.url }</h1><h2>State: ${state}</h2>`);
+     } else if (request.url === "/remove") {
+          state -= 1;
+          response.write(`<h1>You requested the path: ${ request.url }</h1><h2>State: ${state}</h2>`);
+     } else if (request.url === "/reset") {
+          state = 10;
+          response.write(`<h1>You requested the path: ${ request.url }</h1><h2>State: ${state}</h2>`);
+     } else {
+          response.write(`<h1>404 not found</h1>`);
```
jasongwartz: This will still return a status code of 200 - but rather than just send the text '404', we should return an http status code of 404. (https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)
Can you figure out how to do that? The Node HTTP documentation will help you out.

week1/index.js
```diff
@@ -1,38 +1,38 @@
...
-			sendText(response, "File not found")
-		}
-	}
-	
-	response.end()
-})
-
-server.listen(3001)
-
+import HTTP from 'http'
```
jasongwartz: Nice job with these changes. Can you integrate this with the 'state' application, so that the different 'state' pages are served from HTML pages with nice CSS?

week1/HomeWork/index.js
itemah26: Hi Jason, in the last commit I've added a better solution for the notfound pages, can you check it please?

PR #51, opened 2017-10-31
homework-week1

meyadd:


week1/HomeWork/index.js
```diff
@@ -0,0 +1,32 @@
...
+   else if(request.url === "/remove") {
+	   state--;
+	   response.write(String(state));
+   }
+   else if(request.url === "/reset") {
+	   state = 10;
+	   response.write(String(state));
+   }
+   else{
+	   response.write("Not found! , Please check the url path");
```
jasongwartz: Nice job here - for a 'page not found', we probably want to return with an HTTP status code of 404. Can you adjust it so that it returns with a status code of 200 for successful requests, and a 404 for the 'not found' page?

week1/HomeWork/index.js
```diff
@@ -0,0 +1,32 @@
...
+   }
+   
+   response.end()
```
jasongwartz: I notice that you return without declaring a 'Content-Type' header. In this case it's probably fine, but if it were, for example, JSON or HTML, we're leaving it up to the browser to try and figure out what the content was supposed to be. Can you add a Content-Type header, to explicitly declare what kind of content you're returning?

PR #52, opened 2017-11-01
Added homework week1

Elias011:
Added homework week1


week1/index.js
```diff
@@ -1,38 +1,30 @@
...
-			response.setHeader('Location', '/')
-		} else {
-			response.statusCode = 404
-			sendText(response, "File not found")
-		}
-	}
+const http = require('http');
+const port = 3000;
+let state = 10;
+let server = http.createServer((request, response)=>{
```
jasongwartz: Your spacing and formatting is okay (though I prefer a different style), but at least you were consistent with your code. However, the code you copied from me 😛 on line 29 is different - you should try and make sure that your style and formatting is consistent within a file.

For example, on line 4, you write `http.createServer((request, response)=>{` (ie. without spaces around the arrow or brackets).
I wrote `server.listen(port, () => {` on line 29, with spaces around the arrows.

I'd recommend you use spaces around the arrows and curly braces (like, for example, `} else if (request.url === "/remove") {` for line 11), as I think it makes it more readable and is also more consistent with JavaScript conventions. But whichever style you choose, make sure it's always the same.

week1/index.js
```diff
@@ -1,38 +1,30 @@
...
-server.listen(3001)
+	
+	if (request.url === "/add"){
+		state +=1
+		response.write(`<h1>The state is : </h1><h2>${state}</h2>`)
 
-console.log('Server started')
\ No newline at end of file
+	}else if (request.url === "/remove"){
+		state -=1
```
jasongwartz: Again here with the formatting - `state -=1` is a bit hard to read, I'd recommend `state -= 1`. A small thing, but it makes a difference 😄

week1/index.js
```diff
@@ -1,38 +1,30 @@
...
+		response.write(`<h1>The state is : </h1><h2>${state}</h2>`)
```
jasongwartz: Without setting a Content-Type header, we're leaving it up to the browser to try and guess what type this is supposed to be. Can you add a Content-Type header so that it's explicitly declared that this will be html?

(Also note that all of your `response.write()` strings are only partial html, not full html. you should probably return a full html page (with `<html>`, `<head>`, and `<body>`)

week1/index.js
```diff
@@ -1,38 +1,30 @@
...
+	
+	}else if (request.url === "/reset"){
+		state = 10
+		response.write(`<h1>The state is : </h1><h2>${state}</h2>`)
+		
+	}else if (request.url === "/state"){
+		response.write(`<h1>The state is :</h1><h2>${state}</h2>`)
+		
+	}else{
+		response.write(`<h1> Error 404 Invalid Request!</h1>`)
```
jasongwartz: Since you haven't set an http status code, this will probably return 200 (the default). In the case of a 'not found', we probably want to return with a status code of 404 - can you make this change? Check the Node HTTP documentation for details.

Also note that "Invalid Request" is usually associated with status 400, and is different from 404 "Not Found". Check for details: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

Elias011:
Thanks Json for the advices
I will keep them in my mind

PR #53, opened 2017-11-01
project completed

pesh12:
Enes's project

### DIFF
--- a/week1/index.js
+++ b/week1/index.js
@@ -1,38 +1,44 @@
-import HTTP from 'http'
-import Path from 'path'
+//import HTTP from 'http'
+//import Path from 'path'
+//
+//import sendIndexHTML from './responses/sendIndexHTML'
+//import sendPage2HTML from './responses/sendPage2HTML'
+//import sendStylesCSS from './responses/sendStylesCSS'
+//import sendText from './responses/sendText'
 
-import sendIndexHTML from './responses/sendIndexHTML'
-import sendPage2HTML from './responses/sendPage2HTML'
-import sendStylesCSS from './responses/sendStylesCSS'
-import sendText from './responses/sendText'
+const http = require('http');
+const port = 3000
+let state = 10;
 
-const server = HTTP.createServer((request, response) => {
-	console.log(request.method, request.url)
 
-	switch (request.url) {
-	case '/':
-		sendIndexHTML(response)
-		break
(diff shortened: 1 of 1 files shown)


week1/index.js
```diff
@@ -1,38 +1,44 @@
...
+    console.log(`I have received a request for path: ${ request.url }`);
+
+    
+    
+    if ( request.url === "/" ) {
+        response.write("You requested root path.");
+    } else if ( request.url === "/state" ) {
+        response.write(`<h1>${state}</h1>`);
+         
+    }else if ( request.url === "/add" ) {
```
jasongwartz: Careful with your formatting here.

`}else if` is much more difficult to read than `} else if`, and make sure you're consistent about it everywhere.

week1/index.js
```diff
@@ -1,38 +1,44 @@
...
+        state++;
+        response.write(`<h1>${state}</h1>`);
+    }else if ( request.url === "/remove" ) {
+        state--;
+        response.write(`<h1>${state}</h1>`);
+    }else if ( request.url === "/reset" ) {
+        state = 10;
+        response.write(`<h1>${state}</h1>`);
+    }else {
+        response.write(`return error code 404: Not found with a friendly message`)
```
jasongwartz: Since you haven't set an http status code, this will probably return 200 (the default). In the case of a 'not found', we probably want to return with a status code of 404 - can you make this change? Check the Node HTTP documentation for details.

Also, you can add your own friendly message!

week1/index.js
```diff
@@ -1,38 +1,44 @@
...
+const server = http.createServer( (request, response) => {
+
+    console.log(`I have received a request for path: ${ request.url }`);
+
+    
+    
+    if ( request.url === "/" ) {
+        response.write("You requested root path.");
+    } else if ( request.url === "/state" ) {
+        response.write(`<h1>${state}</h1>`);
```
jasongwartz: Without setting a Content-Type header, we're leaving it up to the browser to try and guess what type this is supposed to be. Can you add a Content-Type header so that it's explicitly declared that this will be html?

(Also note that all of your response.write() strings are only partial html, not full html. you should probably return a full html page (with <html>, <head>, and <body>)

pesh12:
Thank you for suggestions Jason. 
I have improved the project according to you suggestions.

PR #57, opened 2017-11-05
homework for NodeJS

aymankanon:
sorry for being late

### DIFF
--- a/week1/index.js
+++ b/week1/index.js
@@ -1,38 +1,27 @@
-import HTTP from 'http'
-import Path from 'path'
 
-import sendIndexHTML from './responses/sendIndexHTML'
-import sendPage2HTML from './responses/sendPage2HTML'
-import sendStylesCSS from './responses/sendStylesCSS'
-import sendText from './responses/sendText'
+const http = require('http');
+const port = 3000
+var state = 10
 
-const server = HTTP.createServer((request, response) => {
-	console.log(request.method, request.url)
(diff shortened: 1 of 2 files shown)


week1/index.js
```diff
@@ -1,38 +1,27 @@
...
+    response.write(String(state))
+}
+else if (request.url === "/remove") {
+    state--;
+    response.write(String(state))
+}
+else if (request.url === "/reset") {
+    state = 10;
+    response.write(String(state))
+}
```
jasongwartz: Very close. But what happens if I request a url other than the ones you have an 'if' condition for?

week1/index.js
```diff
@@ -1,38 +1,27 @@
...
-import sendPage2HTML from './responses/sendPage2HTML'
-import sendStylesCSS from './responses/sendStylesCSS'
-import sendText from './responses/sendText'
+const http = require('http');
+const port = 3000
+var state = 10
 
-const server = HTTP.createServer((request, response) => {
-	console.log(request.method, request.url)
+let server = http.createServer( (request, response) => {
```
jasongwartz: This is good, but doesn't quite cover all the things we talked about in class. If you have time, I'd like to see you implement setting a Content-Type header and returning some HTML content, interpolated with the 'state' value, instead of just returning `String(state)` for every page.

aymankanon:
I am not sure whether it works or not

PR #58, opened 2017-11-05
Week2

pesh12:
Week 2 assignment


week2/index.js
```diff
@@ -0,0 +1,65 @@
...
+    fs.readFile("./list.txt", "UTF-8", function (err, contents) {
+
+        if (err) console.log(err);
+        console.log("\nHere is the list \n" + contents);
+    });
+}
+
+if (process.argv[2] === "add") {
+    let addContent = "\n" + process.argv[3].toString();
+    fs.appendFileSync("./list.txt", addContent);
```
jasongwartz: This is one way, but only works for plain-text files. Can you implement it using only readFile and writeFile?

week2/index.js
```diff
@@ -0,0 +1,65 @@
...
+    fs.readFile("./list.txt", "UTF-8", function (err, contents) {
```
jasongwartz: You need to do this only because you don't already have the full file contents. If you use readFile and writeFile, you might be able to get rid of this readFile.

week2/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+        if (err) console.log(err);
+        console.log("\nHere is the new list\n" + contents);
+    });
+}
+
+if (process.argv[2] === "remove") {
+    let index = +(process.argv[3]) - 1;
```
jasongwartz: haha, that's a 'cheeky' way of doing string-to-integer casting. But what will happen if process.argv[3] is not a valid integer? What if I run `node index.js remove asdf`? (hint - it's really bad)

PR #59, opened 2017-11-06
Ji homework week 2

JiDarwish:
I finished what I started in class so I didn't actually append the text but I overwrote it each time because I felt more comfortable doing that.
I will however try and make another version using the append text thing.
There's really a lot of code and I had some ideas that I though were interested so I added them to as extra.
So a feedback would be really amazing thanks a lot!

PS: I didn't know how to only commit the new homework only so you actually can also see homework from last week I don't know how to fix this now but I'll look it up. Please ignore everything except `week2/index.js` and `week2/help.txt`

### DIFF
--- a/week2/index.js
+++ b/week2/index.js
@@ -1,0 +1,213 @@
+const fs = require('fs');
+const readline = require('readline');
+let todos = [];//to be filled todo list
+
+//check if there is a parameter
+if(process.argv[2] !== undefined){
+    start();
+} else {
+    console.log('Please specify an option to use, here is the manual\n');
+    manualReader('./help.txt');
+}
+
+
+function start(){
+    switch(process.argv[2].trim().toLowerCase()){
+        case 'help':
+            manualReader('./help.txt');
+            break;
+        case 'add':
+            checkFileContent().then((item) => addTodoItem(item)).catch((err) => console.log(err));
+            break;
+        case 'list':
+            console.log('The items on your todo list: ');
+            todoFileReader('./todos.txt');
+            break;
+        case 'remove':
+            itemRemover(process.argv[3]);
+            break;
+        case 'reset':
+            itemsResetter();
+            break;
+        case 'cross':
+            crossItem(process.argv[3]);
+            break;
+        case 'uncross':
+            uncrossItem(process.argv[3]);
+            break;
+        case 'update':
+            update(process.argv[3]);
+            break;
+        default:
+            console.log('No such option! Take a look at the manual\n');
+            manualReader('./help.txt');
+            break;
+    }
+} 
+
+
+
+//  display the manual
+function manualReader(fileName){
+    fs.readFile(fileName, 'UTF-8', (err, data) => {
+        console.log(data);
+    });
+}
+
+//  display the todo file once called
+function todoFileReader(fileName){
+    fs.exists(fileName, (exists) => {
+        //see if file exists to add the existing items to the array
+        if(exists){
+            fs.readFile(fileName, 'UTF-8', (err, data) => {
+                if (data === ""){
+                    console.log('There are no items yet, try using \'node index.js add "some todo item"\'');
+                    return;
+                } 
+                data = data.split('\r\n');
+                let text = data.reduce(((last, item, index) => last += `${index + 1}. ${item}\n`), ``);
+                console.log(text);
+            });
+        } else {
+            console.log('There are no items yet, try using \'node index.js add "some todo item"\'');
+        }
+    });
+}
+
+
+
+//check if there's already a file and push its contents into the todos array
+function checkFileContent(){
+    return new Promise(function(resolve, reject){
+        fs.exists('./todos.txt', (exists)=>{
+            if(!exists){
+                resolve(process.argv[3]);
+                return;                
+            }
+            fs.readFile('todos.txt', 'UTF-8', (err, data) => {
+                todos = data.trim().split('\r\n');
+                resolve(process.argv[3]);
+            });
+        });
+    });
+}
+
+//add the new item to the array
+function addTodoItem(todoItem){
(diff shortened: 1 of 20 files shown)


week2/index.js
```diff
@@ -0,0 +1,213 @@
...
+if(process.argv[2] !== undefined){
+    start();
+} else {
+    console.log('Please specify an option to use, here is the manual\n');
+    manualReader('./help.txt');
+}
+
+
+function start(){
+    switch(process.argv[2].trim().toLowerCase()){
```
jasongwartz: Good use of 'switch' instead of 'if/elseif/elseif/elseif/elseif......'

week2/index.js
```diff
@@ -0,0 +1,213 @@
...
+        case 'help':
+            manualReader('./help.txt');
```
jasongwartz: I like that you break out into another function call here, instead of putting things inline. That makes it easy to see what the different options are, and where to go look for the corresponding code.

week2/index.js
```diff
@@ -0,0 +1,213 @@
...
+//  display the manual
+function manualReader(fileName){
+    fs.readFile(fileName, 'UTF-8', (err, data) => {
+        console.log(data);
+    });
+}
+
+//  display the todo file once called
+function todoFileReader(fileName){
+    fs.exists(fileName, (exists) => {
```
jasongwartz: Good job here - this is a different approach as compared to attempting to read the file and handle the error if the file doesn't exist.

What you've done - checking this first - is usually called 'Look Before You Leap'. https://docs.python.org/3/glossary.html#term-lbyl
The other way - just try and see what happens - is called 'Easier to ask for Forgiveness than Permission'. https://docs.python.org/3/glossary.html#term-eafp

week2/index.js
```diff
@@ -0,0 +1,213 @@
...
+    });
+    console.log('Resetted the todo list');
+}
+
+
+//  modify an item based on filled index
+function update(index){
+    //check if an index was actually filled in
+    if(index !== undefined){
+        index = parseInt(index) - 1;
```
jasongwartz: What if index exists but isn't a valid int? What will happen here? (might be really bad!)

week2/index.js
```diff
@@ -0,0 +1,213 @@
...
+            });
+        });
+    } else {
+        console.log('Not the correct usage! Try "node index.js help"');                
+    }
+    
+}
+
+
+//  mark item as DONE
```
jasongwartz: Great feature!

week2/index.js
```diff
@@ -0,0 +1,213 @@
...
+            index = parseInt(index) - 1;  
+            //check if there's a corresponding todo item for the index and if the item doesn't have the word 'DONE' in it          
+            if(data[index] !== undefined && data[index].indexOf('DONE') === -1){
+                let item = data[index];
+                data[index] += '\t\t\tDONE';//append the word 'DONE'
+                console.log('crossed!');
+                todosWriter(data);
+                return;
+            } 
+            //if the todo item already has the word 'DONE' in it
```
jasongwartz: Interesting technique... I think it works, but is probably not the most sustainable. What if we want to start adding other fields of information other than 'Done', like 'Assignee' and 'Priority'? (see the main comment about JSON)

week2/index.js
JiDarwish: Hahaha I had no idea, the names are so great! 😆

week2/index.js
JiDarwish: Fixed! Good noticed haha I kept looking for these cases for a while but apparently not all of them are gone haha

jasongwartz: I don't know if you addressed my comment with this commit 😛 

Try running `parseInt( 'this is not a number' ) - 1`, and see if the result will cause problems when index is that value.

week2/index.js
JiDarwish: Yes I don't fully understand how I'm going to make what you're talking about work with JSON but I'll try my best to figure it out.

JiDarwish: Thanks a lot for you feedback!!

jasongwartz: Right now, your file looks like this:
```
TodoItem line one.
Another item on line two.
```

You could alternatively use a JSON schema, something like:
```
{
    todos: [
        {
            text: "TodoItem line one.",
            done: true
         },
         {
             text: "Another item on line two.",
             done: false
         }
    ]
}
```

That allows you to store arbitrary data (meaning whatever kinds of information you want) about each todo item.

JiDarwish: Hi Jason, thanks now I get it.
Actually I already improvised something, I don't know if you think it's good or not but it adds more flexibility to the 'cross' thing, now it's called 'mark' hahaha.
Please let me know what you think of it if you had time. Thanks!

week2/v2/index.js
```diff
@@ -0,0 +1,176 @@
...
+        } else{
+            console.log('There\'s no mark! Try "node index.js list" to see the items');            
+        }
+    });    
+}
+
+function updateItem(index){
+    index = parseInt(index) - 1;
+    fs.readFile('./todos.json', 'UTF-8', (err, data) => {
+        data = JSON.parse(data);
```
jasongwartz: Be careful about re-assigning variables. This works in javascript but not in all languages, and can be a bit of a bad practice anyway - it can lead to confusion about what's actually inside that variable. Better to use a new variable name for after it's been parsed.

Either `let data = JSON.parse(rawData)` or `let parsedData = JSON.parse(data)`, something like that.

JiDarwish: I agree, I was wondering myself but then I thought, less names should be better haha. I will modify it.

JiDarwish:
I've modified my homework for the second week based your feedback @jasongwartz which I'm very thankful for. Now I close this PR

PR #60, opened 2017-11-07
Week2 branch

meyadd:
my homework in two ways:
the first deals with items as array of items and saves it as text
the second deals with items as properties of object and saves it as json format


week2/Homework/Json/j_index.js
```diff
@@ -0,0 +1,109 @@
...
+		});
+    });
+}
+function writeFile(file_name , file_data){return new Promise(
+    (resolve, reject) => {
+		fs.writeFile(file_name, file_data, function (err) {
+			  if (err) {
+				reject(err);
+			  }
+			  else{
```
jasongwartz: Be careful about your formatting. It might seem like it doesn't matter (the computer doesn't care, after all), but code is meant to be read by others 😄 

If you're going to write `if ( condition ) {`, make sure you write ` } else {`. ie. if you put spaces in between the keywords and the brackets, always do it.

This is also the case on line 35, where you're written `switch(command){` (the parentheses kind of make it look like a function call, which it's not). I would rather see it written as `switch (command) {`, or even `switch ( command ) {`.

You've also used tabs for indentation, which is no big deal, but take a look at how it appears on github - it seems to be showing 8 spaces per tab, which is a lot, and breaks up your code quite a bit.

week2/Homework/Json/j_index.js
```diff
@@ -0,0 +1,109 @@
...
+		.then((result) => console.log(result))
+		.catch((error) => console.log(error));
+		break;
+	case "list":
+		readFile('j_items.txt')
+		.then((result) => console.log(result))
+		.catch((error) => console.log(error));
+		break;
+	case "add":
+		if (process.argv.length >= 5){
```
jasongwartz: Comments would be helpful here. I see that you're using process.argv.length, but without a lot of effort reading your code, it's not super clear why. What happens if process.argv.length is less than 5? (from your code, it looks like nothing will happen - no error message, just nothing).

I'd like to see something like:
```
case "add":
    // usage: node index.js add 3 'this is the text of my stuff'
    ...code...
```

PR #72, opened 2018-01-15
I made my homework

yousefallaban:
homework week0

### DIFF
--- a/week1/index.js
+++ b/week1/index.js
@@ -1,52 +1,59 @@
-// To run this file: node index.js
-// Then in your browser: http://localhost:3000
 const http = require('http');
-const port = 3000;
-let state = 10;
+const port = 8080;
+//  create a server, and that server will take care of handling requests
+//  for data, and then returning some information.
+//pass back header to let the client know what types of files the server will be returning.
 const server = http.createServer((request, response) => {
 	console.log(request.method, request.url);
-	//console.log(response);
+	response.writeHeader(200, {
+		'content-type': 'text/html'
+	});
 });
-// Start the HTTP server, start listening for request
+// Start the HTTP server, start listening for requests
 server.listen(port, (error) => {
 	if (error) {
 		console.log(error);
 	} else {
-		console.log('api listening on port', port);
+		console.log(`Go to http://localhost:${port} on your browser.`);
 	}
 });
-// Create a event handler for "request"
-// this is an alternative way
+
+server.on("connection", () => {
+	console.log("connected");
+});
+
+
+function htmlMessage(message, response) {
+	response.write(
+		`<html>
+			<head>
+			</head>
+			<body>
+				<h1>${message}</h1>
+			</body>
+		</html>`
+	);
+}
+
+let state = 10;
 server.on('request', (request, response) => {
 	console.log('New http request received', request.url);
 	switch (request.url) {
 		case "/state":
-			response.setHeader('content-type', 'text/html');
-			response.write(`<html><head></head><body><h1>The state :${state}</h1></body></html>`);
-			response.end();
+			htmlMessage(`The state: ${state}`, response);
 			break;
 		case "/add":
-			response.setHeader('content-type', 'text/html');
-			response.write(`<html><head></head><body><h1>Adding: ${state++}</h1></body></html>`);
-			response.end();
+			htmlMessage(`Adding: ${state++}`, response);
 			break;
 		case "/remove":
-			response.setHeader('content-type', 'text/html');
-			response.write(`<html><head></head><body><h1>Removing: ${state--}</h1></body></html>`);
-			response.end();
+			htmlMessage(`Removing: ${state--}`, response);
 			break;
 		case "/reset":
-			response.setHeader('content-type', 'text/html');
-			response.write(`<html><head></head><body><h1>reset: ${state = 10}</h1></body></html>`);
-			response.end();
+			htmlMessage(`Reset: ${state = 10}`, response);
 			break;
 		default:
 			response.statusCode = 404;
-			response.write(`<html><head></head><body><h1>The page not found__${response.statusCode}</h1></body></html>`); response.end();
-			response.end();
-
+			htmlMessage(`The page is not found__404`, response);
 	}
-});
-server.on("connection", () => {
-	console.log("connected");
+	response.end();
 });
--- a/week1/index.js
+++ b/week1/index.js
@@ -1,38 +1,52 @@
-import HTTP from 'http'
-import Path from 'path'
-
-import sendIndexHTML from './responses/sendIndexHTML'
-import sendPage2HTML from './responses/sendPage2HTML'
-import sendStylesCSS from './responses/sendStylesCSS'
-import sendText from './responses/sendText'
-
-const server = HTTP.createServer((request, response) => {
-	console.log(request.method, request.url)
-
-	switch (request.url) {
-	case '/':
-		sendIndexHTML(response)
-		break
-	case '/page2':
-		sendPage2HTML(response)
-		break
-	case '/styles.css':
-		sendStylesCSS(response)
-		break
-	default:
-		const extension = Path.extname(request.url)
-		if (extension === '') {
-			response.statusCode = 302
-			response.setHeader('Location', '/')
-		} else {
-			response.statusCode = 404
-			sendText(response, "File not found")
-		}
+// To run this file: node index.js
+// Then in your browser: http://localhost:3000
+const http = require('http');
+const port = 3000;
+let state = 10;
+const server = http.createServer((request, response) => {
+	console.log(request.method, request.url);
+	//console.log(response);
+});
+// Start the HTTP server, start listening for request
+server.listen(port, (error) => {
+	if (error) {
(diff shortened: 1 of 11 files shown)


week1/index.js
```diff
@@ -1,38 +1,52 @@
...
-			sendText(response, "File not found")
-		}
+// To run this file: node index.js
+// Then in your browser: http://localhost:3000
+const http = require('http');
+const port = 3000;
+let state = 10;
+const server = http.createServer((request, response) => {
+	console.log(request.method, request.url);
+	//console.log(response);
```
jasongwartz: Remove commented-out code before pushing.

week1/index.js
```diff
@@ -1,38 +1,59 @@
-import HTTP from 'http'
-import Path from 'path'
+const http = require('http');
+const port = 8080;
+//  create a server, and that server will take care of handling requests
+//  for data, and then returning some information.
+//pass back header to let the client know what types of files the server will be returning.
+const server = http.createServer((request, response) => {
+	console.log(request.method, request.url);
+	response.writeHeader(200, {
```
jasongwartz: You respond to requests here, but also further down on line 39?

week2/index.js
```diff
@@ -0,0 +1,96 @@
...
+        fs.readFile(filename, "utf8", (err, data) => {
+            if (err) {
+                if (err.code === "ENOENT") {
+                    console.log("Sorry, this file doesn't exist.");
+                } else if (err.code === "EACCES") {
+                    console.log("Sorry, but you don't have permission.");
+                } else {
+                    console.log(` unknown error occurred: ${err.message}`);
+                }
+                process.exit();
```
jasongwartz: You should `reject` the promise here, instead of exiting, and allow the caller to decide what to do with the rejection - exit, or try something else.

week2/index.js
```diff
@@ -0,0 +1,96 @@
...
+            console.log(`Successfully read the file ${filename}`);
+            resolve(data);
+        });
+    });
+};
+const writeTodosFile = (filename, dataAsString) => {
+    return new Promise((resolve, reject) => {
+        fs.writeFile(filename, dataAsString, (err) => {
+            if (err) {
+                if (err.code === "ENOENT") {
```
jasongwartz: Don't copy-paste without thinking - you won't get a file not found error when writing, because writing creates the file. 😄

week2/index.js
```diff
@@ -0,0 +1,96 @@
...
+const addTodoItem = (filename, todoList, itemText) => {
+    todoList.push({
+        message: itemText
+    });
+    writeTodosFile(TODOS_FILENAME, JSON.stringify(myTodoList)).then(() => console.log("Done adding todo."));
+};
+const removeTodoItem = (filename, todoList, itemIndex) => {
+    if (todoList.length > 0) {
+        todoList.splice(itemIndex - 1, 1);
+        writeTodosFile(TODOS_FILENAME, JSON.stringify(myTodoList)).then(() => console.log("Done Removing the task from your TODO list."));
```
jasongwartz: This line is too long and should be split up.

To avoid writing the same comment over again, please read Mo's homework feedback.

week2/index.js
```diff
@@ -0,0 +1,96 @@
...
+const showHelpMenu = (filename) => {
+    readTodosFile(filename).then((data) => {
+        console.log(data);
+    });
+};
+readTodosFile(TODOS_FILENAME).then((data) => {
+    data = data || "[]";
+    myTodoList = JSON.parse(data);
+    const commandArgs = process.argv.slice(2);
+    if (commandArgs[1] === undefined) {
```
jasongwartz: what if my command is `list`? Why are you checking for undef in `commandArgs[1]` here but `commandArgs[0]` below?

week2/index.js
yousefallaban: it,s not about copy-past.  just I do'nt know what is the  err I have to handel. so but the same eer handling. 
you mean like that:  
if (err) {
                reject(err);
            }
            resolve(`The < ${filename}> file has been saved!`);

week2/index.js
yousefallaban: I couldn't found Mo's..
Can you please explain what do you mean by saying split up..
do you mean . I have to put  .then() in new line?

week2/index.js
yousefallaban: You are right..
my idol is < if the user forgot to write something after ADD - REMOVE> 
so i give him the help file

week2/index.js
```diff
@@ -65,17 +65,18 @@ const listTodoItem = (todoList) => {
     }
 };
 const showHelpMenu = (filename) => {
-    readTodosFile(filename).then((data) => {
-        console.log(data);
-    });
+    readTodosFile(filename)
+        .then((data) => {
+            console.log(data)
+                .catch((err) => {
```
jasongwartz: You have a bracket alignment problem here. You missed a bracket, so the code will behave as if you're trying to get the `.catch()` of the return value of `console.log()`, which obviously doesn't make any sense.

You've written:
```
readTodosFile(filename)
        .then((data) => {
            console.log(data)
                .catch((err) => {
```

But you meant to write:
```
readTodosFile(filename)
        .then((data) => {
            console.log(data)
        }).catch((err) => {
```

ie. the `.catch()` chains after the `.then()`.

jasongwartz:
I'd like it if you could use more descriptive commit messages - `update my cli HM` isn't enough information 😄 

Try reading for a great explanation: https://chris.beams.io/posts/git-commit/

That said, great job with the homework this week! 👍

week3/RESTfulAPIs/app.js
```diff
@@ -0,0 +1,136 @@
+'use strict';
+
+const express = require('express');
+const bodyParser = require('body-parser');
+const fs = require('fs');
+const uniqid = require('uniqid');
+
+const app = express();
+const port = process.env.PORT || 3000;
```
jasongwartz: Nice use of an env var (with a default value) for configuration.

week3/RESTfulAPIs/app.js
```diff
@@ -0,0 +1,136 @@
...
+    });
+    writeTodosFile(TODOS_FILENAME, JSON.stringify(myTodoList))
+        .then(() => console.log("Done adding todo."))
+        .catch((err) => {
+            console.log('ERROR:', err);
+        });
+};
+
+const removeTodoItem = (filename, todoList, todoId) => {
+    const index = todoList.map((x) => {
```
jasongwartz: I'm curious why you choose to make things harder by assigning random identifiers instead of using the element's array index as your identifier.

You'll probably be very happy when you learn about "uuids" for databases 😄

week3/RESTfulAPIs/app.js
```diff
@@ -0,0 +1,136 @@
...
+
+app.use(bodyParser.urlencoded({
+    extended: false
+}));
+app.use(bodyParser.json());
+
+const TODOS_FILENAME = './data/todo.json';
+let myTodoList = [];
+//console.log('my to do list  : ',myTodoList);
+app.listen(port, (err) => {
```
jasongwartz: You run your 'app.listen()` here, but you don't create your routes until line 102 which is inside a Promise.then().

Here's a potential problem:
Imagine that the todos file becomes 100GB in size. Now, it takes quite a long time for your `readTodosFile` function to run.
Since you start listening here, but set up the routes AFTER you've read the todo file, there might be a long time where your server is listening but the routes haven't been created (since the todo file takes a long time to read).

You should either put the call to `readTodosFile` somewhere else, like inside the routes (it's important to think about WHEN functions get executed and lines of code get hit), or put your `app.listen` inside that same Promise.then, so that the server doesn't start listening until it's actually ready.

week3/RESTfulAPIs/app.js
```diff
@@ -0,0 +1,136 @@
...
+                res.send(data);
+            }).catch((err) => console.log(err));
+        });
+
+        app.get('/list', (req, res) => {
+            res.send(myTodoList);
+        });
+
+        app.post('/add/:description', (req, res) => {
+            const description = req.params.description;
```
jasongwartz: You should not use the URL path to pass the data of the todo text. You should use the body, eg: `curl -X POST -d '{ "todoText": "pick up apples" }' http://....`, and then use `req.body`. Remember, this only works if you have `bodyParser`.

week3/RESTfulAPIs/app.js
```diff
@@ -0,0 +1,136 @@
...
+            addTodoItem(TODOS_FILENAME, myTodoList, description);
+            res.send('added Task done..');
```
jasongwartz: So - because `addTodoItem` is async (inside, it calls a function that returns a promise), it's possible that there's a problem or error which means the task cannot be added. But since you don't handle async cases here, you will `res.send()` before the task is done being added. That means if there's a problem, you cannot notify the 'user' who sent this request - you already told them it was done before you had finished trying!

You probably want to refactor this so that `addTodoItem` returns a promise, and then change this to:
```
addTodoItem(TODOS_FILENAME, myTodoList, description).then( () => {
  res.send('added Task done..')
}).catch( () => {
  // error handling... 
});
```

PR #73, opened 2018-01-17
Node.js Week1 Homework

malnajar:


GeorgeSapkin:
Why don't you put your code into a `.js` file?

malnajar:
I will ... thanks for your note

week1/MAKEME.md
```diff
@@ -14,30 +14,51 @@ var state = 10;
...
-// /reset
-// Response: "ok" in html format
-// This should set the state back to 10
-http://localhost:8080/reset
-
-// Any other URL
-// Response: return error code 404: Not found with a friendly message
-// and do not change the state variable
-http://localhost:8080/subtract
+'use strict';
```
jasongwartz: I agree, this should be in a javascript file - not a README.md file.

week1/MAKEME.md
```diff
@@ -14,30 +14,51 @@ var state = 10;
...
+
+// To run this file: node index.js
+// Then in your browser: http://localhost:8080
+let http = require('http');
+let port = 8080;
```
jasongwartz: if these things are not going to change while the programming is running, why did you use `let` instead of `const`?

week1/MAKEME.md
```diff
@@ -14,30 +14,51 @@ var state = 10;
...
+        console.log('api listening on port', port);
+    }
+});
+
+// Create a event handler for "request"
+// this is an alternative way
+server.on('request', function (request, response) {
+    console.log(request.url);
+
+    if (request.url === '/state') {
```
jasongwartz: The `if/elseif/else` works...okay...for this - can you think of another way to handle all these different cases?

week1/MAKEME.md
```diff
@@ -14,30 +14,51 @@ var state = 10;
...
+        response.setHeader('content-type', 'text/html');
+        response.write('<h1>'+ state +'</h1>');
```
jasongwartz: You repeat your `response.write` a lot inside every `if` block - could you refactor this to make it a bit cleaner and repeat less?

PR #74, opened 2018-01-17
add week1 homework

chileluk:


week1/chileshe.js
```diff
@@ -0,0 +1,54 @@
...
+const server = http.createServer((req, res) => {
+    console.log("Creating server");
+});
+
+server.on("connection", () => {
+    console.log("Connected")
+});
+
+
+const PORT = 8080;
```
jasongwartz: Good use of `const` for things that aren't going to change, and extra points for using the convention of all-caps for constants.

week1/chileshe.js
```diff
@@ -0,0 +1,54 @@
...
+});
+
+
+let state = 10;
+
+server.on("request", (req, res) => {
+
+    res.setHeader("content-type", "text/html");
+
+    if (req.url === "/state") {
```
jasongwartz: `if/elseif/else` works okay here, but it's not ideal. Can you think of another way to handle all the different cases?

week1/chileshe.js
```diff
@@ -0,0 +1,54 @@
...
+        res.write(`<h1>state = ${state}</h1>`);
+    }
+    else if (req.url === "/add") {
+        res.write("<h1>ok</h1>");
+        state++;
+        res.write(`<h1>state = ${state}</h1>`);
```
jasongwartz: You repeat `res.write` a lot, maybe there's a way to refactor this and make it a bit cleaner?

week1/chileshe.js
```diff
@@ -0,0 +1,64 @@
...
+
+
+let state = 10;
+
+server.on("request", (req, res) => {
+
+    res.setHeader("content-type", "text/html");
+
+
+    function writeState() {
```
jasongwartz: You're re-declaring these functions every time a new request comes in, which is unnecessary - the function definitions are the same every time.

jasongwartz:
Thanks for making the changes I suggested, and good job with this week's homework 👍

week2/index.js
```diff
@@ -0,0 +1,129 @@
...
+            }
+            break;
+    
+        default:
+        displayHelp();       
+    }
+}
+
+
+readTodoFile();
```
jasongwartz: As a more general comment - all of your code is good, but the code flow is quite confusing. It's a combination of the order in which your functions are written, the names of the functions, and what each function does.

I'd prefer something like:

```
function main(command) {
  readTodosFile().then( (data) => {
    executeTask(command);
  }
};
```

Or something like that - where it's clear what each function does, and the order in which they run.

week2/index.js
```diff
@@ -0,0 +1,129 @@
...
+            }
+            else {
+                tasks = tasks.concat(`${command[1]}\n`);
+                console.log(`Adding "${command[1]}"`);
+                writeTodoFile();
+            }
+            break;
+
+        case "remove":    
+            if (!taskNumber || command.length !== 2 || taskNumber > taskList.length - 1 || taskNumber < 0) {
```
jasongwartz: A comment or two would help explain what all these comparisons are for :)

week2/index.js
```diff
@@ -0,0 +1,129 @@
...
+        case "list":
+            !tasks ? console.log("There are no items on the list") : console.log(tasks);
+            break;
+
+        case "add":
+            if (!command[1] || command.length !== 2 || typeof command[1] !== "string") {
+                displayHelp();
+            }
+            else {
+                tasks = tasks.concat(`${command[1]}\n`);
```
jasongwartz: Here you use `command[1]` as if it's the string contents of the task you want to add, but on line 70, you use `command[1]` as a number - how is that?

week2/index.js
```diff
@@ -0,0 +1,129 @@
...
+}
+
+
+
+function handleError(file, err) {
+
+    if (err.code === "ENOENT") {
+        console.log(`${file} does not exist`);
+
+        if (file === todoFile) {
```
jasongwartz: This could be refactored as `if (err.code === "ENOENT" && file === todoFile)`, since you don't do anything with the `else` cases of either.

week3/index.js
```diff
@@ -1,15 +1,119 @@
...
+    app.use(bodyParser.json());
+
+
+    app.get("/list", (req, res) => {
+        readTodosFile().then(data => {
+            if (!data || data === "[]") res.send("There are no items on the list");
+  
+            res.send(data);
+
+        }).catch(err => console.error(err));
```
jasongwartz: Nice job using a `.catch()` to handle promise rejections. One problem - since you only `res.send()` in the successful case, if you encounter an error that triggers the `catch()` and not the `then()`, the response is never sent (and the browser or client will just wait for the request forever, eventually timing out).

One pattern you could use is that in your error cases, you can send back a JSON document like:
```
{ "success": false, "error": "${ err }" }
```

Or something similar - this way your caller gets the information about the error too.

(If you want to get really fancy - you would also want to send back the appropriate response code: 4xx (meaning 400-something) for client errors (malformed request, request for an element that doesn't exist, etc) and 5xx for server errors like a code bug)

week3/index.js
```diff
@@ -1,15 +1,119 @@
...
+  
+            res.send(data);
+
+        }).catch(err => console.error(err));   
+    });
+
+
+    app.post("/add", (req, res) => {
+
+        const newTodo = req.body.newTodo;
```
jasongwartz: Nice job using bodyParser!

week3/index.js
```diff
@@ -1,15 +1,119 @@
...
+const PORT = 4000;
 
-const {list, create, update, remove} = require('./actions')
+app.listen(PORT, () => {
+    console.log(`Server listening on port ${PORT}`);
+});
+
+
+
+function handleRequest() {
```
jasongwartz: I'm curious why you put this in a function instead of just top-level code. It's not a bad thing - but normally the routes are just created in top-level code, since it only needs to happen once (so it doesn't need the repeatability of a function) and generally uses root-scope objects (like `app`).

week3/index.js
```diff
@@ -12,108 +12,108 @@ app.listen(PORT, () => {
...
-    app.get("/list", (req, res) => {
-        readTodosFile().then(data => {
-            if (!data || data === "[]") res.send("There are no items on the list");
-  
-            res.send(data);
-
-        }).catch(err => console.error(err));   
+    readTodosFile().then(data => {
+        if (!data || data === []) {
+            res.json({ message: "There are no items on the list" });
```
jasongwartz: You probably want to `return` here, or put the next line (`res.json(data)`) in the else block, so that they don't both get triggered when the list is empty.

week3/index.js
```diff
@@ -12,108 +12,108 @@ app.listen(PORT, () => {
...
+        }
+        res.json(data);
+    }).catch(err => {
+        res.json({ success: false, error: err.message });
```
jasongwartz: Thanks for fixing this! 😄

PR #75, opened 2018-01-17
week 1 nodejs is added

jalalalwani:
here is the hm for week1 nodejs,

### DIFF
--- a/week2/jalalhm/test2.js
+++ b/week2/jalalhm/test2.js
@@ -98,7 +98,6 @@ function errors(error) {
 }
 
 // Program start
-
 readTodosFile(TODOS_FILENAME).then((data) => {
     data = data || "[]"
     myTodoList = JSON.parse(data)
--- a/week2/jalalhm/test2.js
+++ b/week2/jalalhm/test2.js
@@ -12,8 +12,8 @@ let readTodosFile = (filename) => {
     return new Promise((resolve, reject) => {
         fs.readFile(filename, "utf8", (err, data) => {
             if (err) {
-                throw err
-                reject()
+                
+                reject(err => errors(err))
             }
             //console.log(`Successfully read the file: ${filename}`)
             resolve(data)
@@ -26,7 +26,7 @@ let writeTodosFile = (filename, dataAsString) => {
     return new Promise((resolve, reject) => {
         fs.writeFile(TODOS_FILENAME, dataAsString, (err) => {
             if (err) {
-                errors(err)
+                 reject(err => errors(err))
             };
             //console.log(`Saved todos file: ${filename}`);
             resolve();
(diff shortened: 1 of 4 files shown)


week2/jalalhm/test2.js
```diff
@@ -0,0 +1,122 @@
+"use strict";
+
+const fs = require('fs');
+
+const TODOS_FILENAME = process.env["TODOS_FILENAME"] || "todos.json";
```
jasongwartz: Nice use of an env var with a default value (unless the filename is "0" or another falsey value).

week2/jalalhm/test2.js
```diff
@@ -0,0 +1,122 @@
...
+let myTodoList = []
+
+// Helper function definitions
+
+let readTodosFile = (filename) => {
+    return new Promise((resolve, reject) => {
+        fs.readFile(filename, "utf8", (err, data) => {
+            if (err) {
+                throw err
+                reject()
```
jasongwartz: It might make more sense here to `reject()` with the error and let the caller handle it, rather than throwing the error and then rejecting the promise.

week2/jalalhm/test2.js
```diff
@@ -0,0 +1,122 @@
...
+
+        })
+    })
+};
+
+let writeTodosFile = (filename, dataAsString) => {
+    return new Promise((resolve, reject) => {
+        fs.writeFile(TODOS_FILENAME, dataAsString, (err) => {
+            if (err) {
+                errors(err)
```
jasongwartz: You should reject the promise here.

week2/jalalhm/test2.js
```diff
@@ -0,0 +1,122 @@
...
+            resolve();
+        })
+    })
+}
+
+let addTodoItem = (filename, todoList, itemText) => {
+    todoList.push({
+        message: itemText
+    })
+    writeTodosFile(filename, JSON.stringify(myTodoList)).then(() => console.log("Done adding todo."));
```
jasongwartz: You need a `.catch()` in case an error occurs or the promise fails (this is true of all of your promise calls that use a `.then()`).

week2/jalalhm/test2.js
```diff
@@ -0,0 +1,122 @@
...
+let removeTodoItem = (filename, index) => {
+    let newMyTodoList;
+    if (index < myTodoList.length) {
+
+        myTodoList.splice(index - 1, 1);
+
+    } else {
+        console.log("Out of bounds");
+    }
+    writeTodosFile(filename, JSON.stringify(myTodoList)).then(() => console.log("element deleted"))
```
jasongwartz: You don't need to write the file if the index was out of bounds.

PR #76, opened 2018-01-17
Added Abdullah

abdullahWeb:


week1/index.js
```diff
@@ -1,38 +1,97 @@
...
+});
+
+server.on('request', (req, res) => {
+
+	console.log("request", req.url);
+
+	let state = 10;
+
+	if (req.url === "") {
+
```
jasongwartz: You have a lot of extra whitespace which can make it difficult to read.

Whitespace is free, so you should use it lots to make your ideas clear - but I think you're using it a bit too much. You're using two or three newlines in a row where you don't need to, and you probably don't need an empty line in between almost every line of code.

week1/index.js
```diff
@@ -1,38 +1,97 @@
...
+
+		req.url = window.location = "http://username.wix.com/sitename#/!page-url/123456";
```
jasongwartz: I think you might be confused about what Node.js is and how it works - `window.location` is undefined in the Node.js runtime because it's a property published by the browser.

week1/index.js
```diff
@@ -1,38 +1,97 @@
...
-server.listen(3001)
 
-console.log('Server started')
\ No newline at end of file
+	res.end();
+
+
+});
+
+const PORT = 4000;
```
jasongwartz: Good use of `const` and uppercase letters for a constant value that doesn't change during the running of the application.

jasongwartz:
You should work on writing better commit messages, so each commit has a description of what it changes. Read this: https://chris.beams.io/posts/git-commit/

week2/index.js
```diff
@@ -0,0 +1,119 @@
+Added Abdullah
```
jasongwartz: Why do you have this line at the top?

week2/index.js
```diff
@@ -0,0 +1,119 @@
...
+        }
+        else if (err.code === "EACCES") {
+            console.log("That file exists but you don't have permission.");
+
+        }
+        else {
+
+            console.log(`Uh oh, an unknown error occured ${err.message}`)
+        }
+        process.exit();
```
jasongwartz: You should reject the promise here (instead of exiting) and let the caller decide how to handle the error through the `.catch()` block.

week2/index.js
```diff
@@ -0,0 +1,119 @@
...
+
+    }
+
+    // console.log(`Successfully read the file: ${TODOS_FILENAME}`)
+
+    myTodoList = JSON.parse(data);
+
```
jasongwartz: I appreciate using whitespace for readability, but because you have an extra line in between every line of code (and sometimes two), it actually makes it harder to read because less code fits on the screen.

Use whitespace for readability, but use it sensibly 😄

week2/index.js
```diff
@@ -0,0 +1,119 @@
...
+    if (commandArgs[0] === "add") {
+
+        addItem();
```
jasongwartz: A better pattern would be to create functions that take a parameter so that you call them with `addItem(itemAsString)`, for example. That way the function doesn't rely on global variables.

week2/index.js
abdullahWeb: I added it when i was editing it from VIM editor in gitbash, but there is no need for it

week2/index.js
abdullahWeb: i thought that we dont need to use promises to solve the homework that's why i didn't use it at all, but i will try to do it again using promises

week2/index.js
abdullahWeb: hahah that is my biggiest problem, i will try to avoid doing that later

week2/index.js
abdullahWeb: Aha, thanks good to know that.

PR #78, opened 2018-01-17
homework week 1

talalalamdar:


week1/index.js
```diff
@@ -1,38 +1,31 @@
...
\ No newline at end of file
+server.on("request", (request, response) => {
+    console.log("requesting", request.url)
+    function setResponse(header) {
+        response.setHeader("Content-Type", "Text-html");
+        response.write(header);
+        response.end();
+    }
+    switch (request.url) {
+        case "/add" : state++; setResponse(`<h1>${state}</h1>`); break;
```
jasongwartz: While I am usually a fan of 'terse' code, this is verging on difficult to read.
In-line case statements are useful when the case does only one thing, but in your case, you have two statements in here (`case "/add": state++; setResponse...`).

However, you might want to explore how to use expressions inside string templates, like:

    setResponse(`<h1>${ state++ }</h1>`);

week1/index.js
```diff
@@ -1,38 +1,31 @@
...
+        case "/remove": state--; setResponse(`<h1>${state}</h1>`); break;
+        case "/reset": state = 10; setResponse(`<h1>${state}</h1>`); break;
+        case "/state": state; setResponse(`<h1>${state}</h1>`); break;
+        default: setResponse(`<h1>Error 404, page not found</h1>`); break;    
+    }
+})
+const port = 8080;
```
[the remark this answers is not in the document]
talalalamdar: Thanks for the feedback Jason, i have updated my code

week1/index.js
```diff
@@ -14,15 +14,15 @@ server.on("request", (request, response) => {
...
     }
     switch (request.url) {
-        case "/add" : state++; setResponse(`<h1>${state}</h1>`); break;
-        case "/remove": state--; setResponse(`<h1>${state}</h1>`); break;
-        case "/reset": state = 10; setResponse(`<h1>${state}</h1>`); break;
-        case "/state": state; setResponse(`<h1>${state}</h1>`); break;
-        default: setResponse(`<h1>Error 404, page not found</h1>`); break;    
+        case "/add" :  setResponse(state++); break;
+        case "/remove":  setResponse(state--); break;
+        case "/reset": setResponse(state = 10); break;
```
jasongwartz: While this might technically work, assigning a variable inside a parameter of a function call is poor style - it's not very clear.

jasongwartz:
Please make your commit messages more descriptive than "modefying the code". I suggest reading:
https://chris.beams.io/posts/git-commit/

jasongwartz:
That said - good job with the homework!

week2/helpMenu.txt
```diff
@@ -0,0 +1,35 @@
+
+    HELP MENU                                                                                    ||
```
jasongwartz: I like the way your help menu is nice and thorough!

There's a bit of a pattern for how to design CLI help menus - I wouldn't have expected you to know this, but you can check some examples for inspiration.

Try `git help`, `curl --help`, `cd --help`, `vim --help`, etc.

week2/index.js
```diff
@@ -0,0 +1,128 @@
...
+                    console.log("I'm sorry, friend, but that file doesn't exist.");
+                } else if (err.code === "EACCES") {
+                    console.log("That file exists but you don't have permission.");
+                } else {
+                    console.log(`Uh oh, an unknown error occured: ${err.message}`);
+                }
+                process.exit()
+            }
+            //  console.log(`Successfully read the file: ${filename}`)
+            resolve(data)
```
jasongwartz: Your promise doesn't reject ever.

(I know this is pretty much unchanged from my example - but I wasn't doing the homework for you, just giving you a starting point)

week2/index.js
```diff
@@ -0,0 +1,128 @@
...
+        break;
+    case "update":
+        updateFunction(commandArgs[1], commandArgs[2])
+        break;
+    case "list":
+        listingTasks();
+        break;
+    case "remove":
+        if (commandArgs[1] == null) {
+            console.log("Task number is undefined.\nCheck the help menu");
```
jasongwartz: I like that you handle the number being undefined here, but comparing to null isn't the best way. What if the number is a string, or 0, or a negative number? You should also do this for the other input functions that take an index, like "update".

week2/index.js
```diff
@@ -0,0 +1,128 @@
...
+        } else {
+            console.log("Tasks to do:")
+            for (let i = 0; i < myTodoList.length; i++) {
+                console.log(`${i + 1}- ${myTodoList[i].task}`)
+            }
+        }
+    })
+}
+let removeTask = (index) => {
+    readFileFunction(TODOS_FILENAME).then(data => {
```
jasongwartz: You don't have a `.catch()` to handle if this promise is rejected.

week2/index.js
```diff
@@ -0,0 +1,128 @@
...
+            //  console.log(`Successfully read the file: ${filename}`)
+            resolve(data)
+        })
+    })
+};
+
+let writeFileFunction = (filename, dataAsString) => {
+    return new Promise((resolve, reject) => {
+        fs.writeFile(filename, dataAsString, (err) => {
+            if (err) reject(console.log(`error happened while writing the file \n${err.message}`));
```
jasongwartz: In this case, your reject is called with `undefined`.

Why? Because the return value of the parameter it's being passed (console.log) is undefined - console.log prints but doesn't return anything.

You probably want to `reject("Error message")`, and then have the caller console.log it in their `.catch()`.

week2/index.js
```diff
@@ -0,0 +1,128 @@
...
+}
+
+let showHelpMenu = () => {
+    readFileFunction("helpMenu.txt").then((data) => console.log(data))
+}
+
+let resetFunction = () => {
+    myTodoList = [];
+    writeFileFunction(TODOS_FILENAME, "")
+    console.log("The tasks list has been reset")
```
jasongwartz: since `writeFileFunction` returns a promise, your `console.log` will be called before the file has finished writing. You probably want to use `.then()` and `.catch()` here.

week2/index.js
talalalamdar: i didnt know that every promise you need to reject it or catch, thanks for the feedback, i updated my code, i hope it is okay like this

week3/helpMenu.txt
```diff
@@ -0,0 +1,63 @@
...
+---------------------------------------                                                          ||
+$ curl -X GET 'http://localhost:8080/' >>>>>>>>>>> To display the help menu                      ||
+---------------------------------------                                                          ||
+                                                                                                 ||
+---------------------------------------                                                          ||
+$ curl -X GET 'http://localhost:8080/help' >>>>>>>>>>> To display the help menu                  ||
+---------------------------------------                                                          ||
+                                                                                                 ||
+-----------------------------------------------------                                            ||
+$ curl -X POST 'http://localhost:8080/add/{newTask}' >>>>>>>>>>> To add a task to the list,      ||
```
jasongwartz: This is interesting, but not a normal way to pass data in a POST request. The URL path is used to specify location - for example, if we had `todos` but also `calendarEvents`, we could use the path to differentiate.
You should pass the data of the task to add in the "body" of the post request (and then use `bodyParser` to get it out of the body inside your `app.post()` handler.

week3/helpMenu.txt
```diff
@@ -0,0 +1,63 @@
...
+-----------------------------------------------------            where {newTask} without spaces. ||
+                                                                                                 ||
+-----------------------------------------------------                                            ||
+$ curl -X DELETE 'http://localhost:8080/delete/{taskId}' >>>>>>>>>>> To remove a task from the   ||
```
jasongwartz: Now, this is actually okay, sort of. You can read up on REST style - REST is an overused term, but what it actually means is an organised way to structure the URL paths. In REST style, a deletion would actually be:

`curl -X DELETE 'http://localhost:8080/todo/{taskId}`

We don't need the path to include `delete`, because we have that information already from the HTTP verb.

week3/helpMenu.txt
```diff
@@ -0,0 +1,63 @@
...
+-----------------------------------------------------             list, with a passed id         ||
+                                                                  {taskId}.                      ||
+--------------------------------------------                                                     ||
+$ curl -X PUT 'http://localhost:8080/reset' >>>>>>>>>>> To remove all the tasks.                 ||
```
jasongwartz: I would use DELETE instead of PUT for the reset endpoint, since it's a destructive operation.

week3/helpMenu.txt
```diff
@@ -0,0 +1,63 @@
...
+--------------------------------------------                                                     ||
+                                                                                                 ||
+----------------------------------------------------------------                                 ||
+$ curl -X PUT 'http://localhost:8080/update/{taskId}/{newTask}' >>>>>>>>>>> to update a task.    ||
```
jasongwartz: Yes, 'update' is the right place to use PUT.

week3/helpMenu.txt
```diff
@@ -0,0 +1,63 @@
...
+$ node index.js update {n} "updated task" >>>>>>>>> To update task number {n} with new task.     ||
+------------------------------------------                                                       ||
+                                                                                                 ||
+---------------------                                                                            ||
+$ node index.js help >>>>>>>>> To display the help menu, this instruction menu.                  ||
+---------------------                                                                            ||
+                                                                                                 ||
+                                                                                                 || 
+                                                                                                 ||
+==================================================================================================
```
jasongwartz: Thanks for writing a comprehensive 'help' file, it makes it A LOT easier to review and give feedback!

week3/helperFunctions.js
```diff
@@ -0,0 +1,167 @@
...
+            .catch(err => {
+                console.log(`an error happened while reading the file ${err}`)
+                reject(err)
+            });
+    })
+
+};
+
+
+module.exports = {
```
jasongwartz: Nice job reusing code from last week 😄

week3/index.js
```diff
@@ -1,15 +1,77 @@
-const Express = require('express')
-const bodyParser = require('body-parser')
+"use strict"
+const express = require("express");
+const app = express();
+const bodyParser = require("body-parser");
+const fs = require("fs");
```
jasongwartz: You probably don't need to import `fs` here, since it's not used here (that I can see). You only need to `require` a library in the file where it gets used.

week3/index.js
```diff
@@ -1,15 +1,77 @@
...
+
+app.get("/list", (req, res) => {
+    helperFunctions.readFileFunction("to-do-list.json")
+        .then((data) => {
+            data = data || "[]";
+            helperFunctions.myTodoList = JSON.parse(data);
+            if (helperFunctions.myTodoList.length === 0) {
+                res.send("You have no tasks to do, you are free\n");
+            } else {
+                res.send("Tasks to do: " + JSON.stringify(helperFunctions.myTodoList) + "\n");
```
jasongwartz: This is a bit of a problem. The entire document needs to be encoded as JSON, otherwise programs that try and decode it will fail (try doing a request in Postman to see - it won't be able to nicely format it as JSON).
In this case, it's okay to not send back a helpful message because you're dealing only in data - you're not really directly interacting with a user. The 'user' is another program that's requesting data only.
I would simply do `res.send(JSON.stringify( helperFunctions.myTodoList )`, but make sure to set the `Content-Type` header so that the 'client' program knows what type of data it's getting back.

week3/index.js
```diff
@@ -1,15 +1,77 @@
...
+            res.send("A task has been added \n")
+        })
+        .catch(err => res.send("failed to add the task " + err.message + "\n"))
+})
+
+app.put("/update/:id/:updatedTask", (req, res) => {
+    let taskId = req.params.id;
+    let newTask = req.params.updatedTask;
+    helperFunctions.updateFunction(taskId, newTask)
+        .then(() => res.send("Task has been updated\n"))
```
jasongwartz: Nice job with these functions!

A common pattern you'll see for APIs that deal mostly with JSON is to send back a short JSON response, rather than a string like you're doing here. This can be simple, like:
```
{ "result": "success" }
```
And then if there's an error, you can send back:
```
{ "result": "error", "errorMessage": err.message }
```

PR #81, opened 2018-01-18
the assigment is already uploaded

moesi88:


week1/week1-assigment/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "server",
+  "version": "1.0.0",
+  "description": "node.js example \u001b[D\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D\u001b[E\u001b[C\u001b[C\u001b[C\u001b[C\u001b[C\u001b[C HTTP server",
```
jasongwartz: I don't mind you committing your package.json file, but be careful about how your 'description' seems to include a bunch of binary characters or something.

week1/week1-assigment/index.js
```diff
@@ -0,0 +1,52 @@
...
+        console.log('api listening on port', port);
+    }
+});
+
+// Create a event handler for "request"
+// this is an alternative way
+server.on('request', function (request, response) {
+
+
+    if (request.url === "/state") {
```
jasongwartz: `if/elseif/else` works...okay...for this, but can you think of a better way to handle all the different cases?

week1/week1-assigment/index.js
```diff
@@ -0,0 +1,52 @@
...
+        response.setHeader('content-type', 'text/html');
```
jasongwartz: you have `response.setHeader` and `response.end` the same in all the `if` cases, and `response.write` almost the same. Can you think of a way to refactor this to avoid so much repetition?

week1/week1-assigment/index.js
```diff
@@ -0,0 +1,52 @@
+// To run this file: node index.js
+// Then in your browser: http://localhost:8080
+var http = require('http');
+
+var port = 8080;
```
jasongwartz: since these things don't change during the running of the application, `const` would be better than `var` or `let`. You should review the differences between `var`, `let`, and `const`.

moesi88: i knw to difference bt i thought like we should not change those few lines

week1/week1-assigment/index.js
moesi88: switch you mean is better for cases right ?
i already check it out bt there is problem and am still tryna figure out it

week1/week1-assigment/index.js
jasongwartz: Okay, that's understandable - I just wanted to make sure you were aware. We'll look at changing it in the original, thanks for pointing that out.

week1/week1-assigment/index.js
moesi88: function that generate this stuff is better right ?
less logic better you right

week1/week1-assigment/index.js
jasongwartz: Yes, `switch` is what I have in mind. Good luck 😄

week1/week1-assigment/index.js
jasongwartz: Yes, a helper function would be a good idea (you can always look at the other PRs to see how the other students solved this problem)

moesi88: iwill check before i give up hahah iwill try to solve it my self first

week1/week1-assigment/package.json
moesi88: okeeez

week2/index.js
```diff
@@ -0,0 +1,135 @@
...
+            break;
+        default:
+            console.log('No such option! Take a look at the manual\n');
+            manualReader('./help.txt');
+            break;
+    }
+}
+
+
+function manualReader(fileName) {
```
jasongwartz: You could extract your two reader functions a bit better by having one reader function and just calling a callback with the data or returning a promise that resolves with the data, rather than having two functions doing similar things. This way you can let the 'caller' decide what to do with the resulting data, not have the code that comes after reading the file inside these functions.

This better follows the pattern of 'utility functions', where your functions have one key job and they do only that one thing.

week2/index.js
```diff
@@ -0,0 +1,135 @@
...
+
+
+//check if there's already a file and to push the item in
+
+function checkFileContent() {
+    return new Promise(function (resolve, reject) {
+        fs.exists('./todos.txt', (exists) => {
+            if (!exists) {
+                resolve(process.argv[3]);
+                return;
```
jasongwartz: you don't need to 'return' inside a Promise, you just 'resolve' and that's the end.

week2/index.js
```diff
@@ -0,0 +1,135 @@
...
+            }
+            fs.readFile('todos.txt', 'UTF-8', (err, data) => {
+                todos = data.trim().split('\r\n');
```
jasongwartz: Your use of global variables is a bit awkward here. You generally don't want functions to manipulate or change variables that exist outside their scope (in a higher scope). It'd be better to resolve the promise with the parsed data.

week2/index.js
```diff
@@ -0,0 +1,135 @@
...
+                resolve(process.argv[3]);
+            });
+        });
+    });
+}
+
+//add the new item to the array
+function addTodoItem(todoItem) {
+    if (todoItem !== '') {
+        if (todoItem.trim() !== '') {
```
jasongwartz: Is this to check that the string isn't just whitespace?

First of all, you could reduce it to `if ( todoItem !== '' && todoItem.trim() !== '' )`. But since an empty string will still be an empty string after `trim()`, you could just leave out the first check and make it `if ( todoItem.trim() !== '' )`.

But you also need to trim the string again (below on line 90), so you're running the `trim()` function twice. Now, this isn't a big deal because trim is fast on a short string. But what if the string is 100M? Running `trim()` could become rather slow.
You could simplify this with doing the `trim` once (before the `if`) conditional, and then checking and reusing it, like:
```
let trimmedItem = todoItem.trim();
if ( trimmedItem !== '' ) {
  todos.push(trimmedItem);
  ...
}
```

week2/index.js
```diff
@@ -0,0 +1,135 @@
...
+            todosWriter(todos);
+        }
+    } else {
+        console.log('Not the correct usage! Try "node index.js help"');
+    }
+}
+
+// rewrite existing file with new todos array 
+function todosWriter(itemsArr) {
+    itemsArr.filter((a) => a !== '');
```
jasongwartz: Is there a particular case you're hoping to catch here? Or maybe a bug you encountered that this fixes?
First of all, you check for empty strings earlier before you add items, so we're already pretty sure there won't be an empty string here.
But the other problem is that `Array.prototype.filter()` returns the new array, it doesn't modify the original. Since you don't save the output of this `filter()` call, you aren't actually using the `filter()` at all.

week2/index.js
moesi88: okez boss

week2/index.js
moesi88: i dont knw why i still dont get the promiss consept as wll thts why i have alot of mistakes in it -__-

week2/index.js
jasongwartz: Correction (as pointed out to me by @remarcmij, thanks!) - what I said is not quite true.

The normal behaviour is to `resolve()` and then let the function end - but it doesn't force the function to exit like `return` does. So the promise will get resolved (and the caller code inside the `.then()` will execute), but the rest of the code inside the promise will still run.

I would refactor what you wrote here into an `if/else`, instead of returning inside the `if` and having a continuation otherwise - but that's a style judgement, since it seems like this code does actually work the way you intended.

PR #82, opened 2018-01-19
HW commit

JawharBairakdar:
Hi Jason,
how are you i did try to stay with the import method but I don't know that it's not working so what i did is copping the files content and put it inside the file it self.
and i did push my homework that i done in the class. :)

### DIFF
--- a/week1/My-own-modified/index.js
+++ b/week1/My-own-modified/index.js
@@ -1,6 +1,5 @@
 "use strict";
 const http = require("http");
-// console.log(http.STATUS_CODES)
 const server = http.createServer((req, res) => {
   console.log("Creating Server");
 });
--- a/week1/My-own-modified/index.js
+++ b/week1/My-own-modified/index.js
@@ -1,0 +1,43 @@
+"use strict";
+const http = require("http");
+const server = http.createServer((req, res) => {
+  console.log("Creating Server");
+});
+let state = 10;
+let original = 10;
+server.on("request", (req, res) => {
+  console.log(req.method, req.url);
+  switch (req.url) {
+    case "/state":
+      res.setHeader("content-type", "text/html");
+      res.write(`<h1>${state}</h1>`);
+      res.end()
+      break;
+    case "/add":
+      res.setHeader("content-type", "text/html");
+      res.write(`<h1>${state += 1}</h1>`);
+      res.end()
+      break;
+    case "/remove":
+      res.setHeader("content-type", "text/html");
+      res.write(`<h1>${state -= 1}</h1>`);
+      res.end()
+      break;
+    case "/reset":
+      res.setHeader("content-type", "text/html");
+      res.write(`<h1>${original}</h1>`);
+      res.end()
+      break;
+    default:
+      res.statusCode = 404
+      res.write(`<h1>${res.statusCode}</h1>`);
+      res.end()  
+  }
+});
+server.on("connection", () => {
+  console.log("connected")
+});
+const PORT = 4000;
+server.listen(PORT, () => {
+  console.log("lestening to", PORT)
+})
--- a/week1/My-own-modified/index.js
+++ b/week1/My-own-modified/index.js
@@ -4,31 +4,32 @@ const server = http.createServer((req, res) => {
   console.log("Creating Server");
 });
 let state = 10;
-let original = 10;
+const ORIGINAL = 10;
+function responseWrite(val, response) {
+  response.setHeader("content-type", "text/html");
+  response.write(`<h1>${val}</h1>`);
+  response.end()
+}
 server.on("request", (req, res) => {
-  function responseWrite(val) {
-    res.setHeader("content-type", "text/html");
-    res.write(`<h1>${val}</h1>`);
-  }
   console.log(req.method, req.url);
   switch (req.url) {
     case "/state":
-      responseWrite(state)
+      responseWrite(state, res)
       break;
     case "/add":
-      responseWrite(state += 1);
+      responseWrite(state += 1, res);
       break;
     case "/remove":
-      responseWrite(state -= 1);
+      responseWrite(state -= 1, res);
       break;
     case "/reset":
-      responseWrite(original);
+      state = ORIGINAL  
+      responseWrite(state, res);
     break;
     default:
       res.statusCode = 404
-      responseWrite(res.statusCode);
+      responseWrite(res.statusCode, res);
   }
-  res.end()  
 });
 server.on("connection", () => {
   console.log("connected")
--- a/week1/My-own-modified/index.js
+++ b/week1/My-own-modified/index.js
@@ -1,5 +1,6 @@
 "use strict";
 const http = require("http");
+// console.log(http.STATUS_CODES)
 const server = http.createServer((req, res) => {
   console.log("Creating Server");
 });
--- a/week1/My-own-modified/index.js
+++ b/week1/My-own-modified/index.js
@@ -6,33 +6,29 @@ const server = http.createServer((req, res) => {
 let state = 10;
 let original = 10;
 server.on("request", (req, res) => {
+  function responseWrite(val) {
+    res.setHeader("content-type", "text/html");
+    res.write(`<h1>${val}</h1>`);
+  }
   console.log(req.method, req.url);
   switch (req.url) {
     case "/state":
-      res.setHeader("content-type", "text/html");
-      res.write(`<h1>${state}</h1>`);
-      res.end()
+      responseWrite(state)
       break;
     case "/add":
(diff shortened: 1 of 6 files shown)


week1/My-own-modified/index.js
```diff
@@ -0,0 +1,43 @@
...
+    default:
+      res.statusCode = 404
+      res.write(`<h1>${res.statusCode}</h1>`);
+      res.end()  
+  }
+});
+server.on("connection", () => {
+  console.log("connected")
+});
+const PORT = 4000;
```
jasongwartz: Good use of const and capitals for things that won't change

week1/My-own-modified/index.js
```diff
@@ -0,0 +1,43 @@
...
+});
+let state = 10;
+let original = 10;
+server.on("request", (req, res) => {
+  console.log(req.method, req.url);
+  switch (req.url) {
+    case "/state":
+      res.setHeader("content-type", "text/html");
+      res.write(`<h1>${state}</h1>`);
+      res.end()
```
jasongwartz: You call `res.setHeader`, `res.write`, and `res.end` in almost every one of these cases. How would you refactor this to avoid so much duplication?

week1/My-own-modified/index.js
```diff
@@ -0,0 +1,43 @@
+"use strict";
+const http = require("http");
+const server = http.createServer((req, res) => {
+  console.log("Creating Server");
+});
+let state = 10;
+let original = 10;
+server.on("request", (req, res) => {
+  console.log(req.method, req.url);
+  switch (req.url) {
```
jasongwartz: Good idea to use a switch statement here.

week1/My-own-modified/index.js
JawharBairakdar: Done.

week1/My-own-modified/index.js
```diff
@@ -0,0 +1,39 @@
...
+      responseWrite(state)
+      break;
+    case "/add":
+      responseWrite(state += 1);
+      break;
+    case "/remove":
+      responseWrite(state -= 1);
+      break;
+    case "/reset":
+      responseWrite(original);
```
jasongwartz: I like the way you structured your refactoring, but line 25 (here) doesn't reset the state to the original, it just prints the original.

week1/My-own-modified/index.js
```diff
@@ -0,0 +1,39 @@
+"use strict";
+const http = require("http");
+const server = http.createServer((req, res) => {
+  console.log("Creating Server");
+});
+let state = 10;
+let original = 10;
```
jasongwartz: since the 'original' doesn't change, it could be a const.

week1/My-own-modified/index.js
```diff
@@ -0,0 +1,39 @@
...
+server.on("request", (req, res) => {
+  function responseWrite(val) {
```
jasongwartz: I really like your use of a helper function, but in this code here, you're re-declaring this function on every request - no need to do that. You could move this outside the `server.on` block.

week1/My-own-modified/index.js
```diff
@@ -0,0 +1,39 @@
...
+      responseWrite(state -= 1);
+      break;
+    case "/reset":
+      responseWrite(original);
+    break;
+    default:
+      res.statusCode = 404
+      responseWrite(res.statusCode);
+  }
+  res.end()
```
jasongwartz: `res.end()` could be moved inside the helper function.

jasongwartz:
Nice work on the fixes, looks good to me.

week2/new Program/index.js
```diff
@@ -0,0 +1,108 @@
+"use strict"
+/****  it's a program that do the following:
```
jasongwartz: Thanks for adding a spec, so a reviewer/reader can see what the code is supposed to do 😄

week2/new Program/index.js
```diff
@@ -0,0 +1,108 @@
...
+  - help: show help section (node index.js help)
+  - list: show current todo's, or show an appropriate text if there are no todos (node index.js list)
+  - add: add a todo item. all the words behind "add" are entered as 1 todo item to the list (node index.js add "Buy groceries")
+  - remove: remove a todo item by its 1-base index. (node index.js remove 2)
+  - reset: remove all todo items from the list (node index.js reset)
+  - BONUS: update: update a todo item with new text (node index.js update 3 "Wash teeth")
+*/
+const FS = require("fs")
+const REQUIRE_ARGS = process.argv.slice(2)
+const FIRST_INPUT_COMMAND = REQUIRE_ARGS[0]
```
jasongwartz: Constants with capital letters aren't usually used for user input. It typically signifies that there's a "magic value" somewhere in the code, so it would be changed by changing code somewhere - but because it's a constant, we can reuse the constant everywhere instead of having to copy the magic value.

You can still use `const`, since it's not going to change while the program is running - but I wanted to explain a bit about the idiom of the capital letters thing, since it looks weird to me here.

week2/new Program/index.js
```diff
@@ -0,0 +1,108 @@
...
+The following commands should be present:
+  - No command: show help section (node index.js)
+  - help: show help section (node index.js help)
+  - list: show current todo's, or show an appropriate text if there are no todos (node index.js list)
+  - add: add a todo item. all the words behind "add" are entered as 1 todo item to the list (node index.js add "Buy groceries")
+  - remove: remove a todo item by its 1-base index. (node index.js remove 2)
+  - reset: remove all todo items from the list (node index.js reset)
+  - BONUS: update: update a todo item with new text (node index.js update 3 "Wash teeth")
+*/
+const FS = require("fs")
```
jasongwartz: It's idiomatic to make the variable/const name the same as the library name, even though that means using lowercase letters.

week2/new Program/index.js
```diff
@@ -0,0 +1,108 @@
...
+} // filesReader()
+
+function filesWriter(path, newData, oldData) {
+  return new Promise((resolve, reject) => {
+    
+    let dataHolder = []
+    if (!oldData) {
+      dataHolder.push(newData)
+    } else if (!newData && !oldData) {
+      dataHolder.push(newData)
```
jasongwartz: Since this line requires `!oldData`, but that's also the condition for line 37, line 40 will never get called. You can just remove it, since line 37 covers all the same situations.

week2/new Program/index.js
```diff
@@ -0,0 +1,108 @@
...
+    })
+  })
+}
+
+function readingError(e) {
+  // this Function is if anything went wrong during the reading mode
+  //   it have to show us the message that explain the error.
+  //   And if the error was "File not found" it will create it automaticly
+  if (e.code === "ENOENT") {
+    filesWriter(FILE_NAME).catch(e => { console.error(e) })
```
jasongwartz: Since the error was that the file doesn't exist, you do the right thing to just create it here. But then on line 64, you exit - even though, for this particular case, you handled it. So you don't need to exit, you could return out here and carry on with the rest of the program.

week2/new Program/index.js
JawharBairakdar: looking forward For that 🙂

week2/new Program/index.js
```diff
@@ -0,0 +1,106 @@
...
+    return
```
jasongwartz: Haha well it works like this - but my idea was more that if you encountered the error and it didn't work you could then actually try the command again (because you know why it failed the first time, and you've fixed it).

In your case it's not so easy, because your switch statement isn't in a function - but if it was, it'd simply be a matter of calling the function again.

week2/new Program/index.js
```diff
@@ -34,9 +34,7 @@ function filesWriter(path, newData, oldData) {
   return new Promise((resolve, reject) => {
     
     let dataHolder = []
-    if (!oldData) {
-      dataHolder.push(newData)
-    } else if (!newData && !oldData) {
+    if (!oldData && newData || !newData && !oldData) {
```
jasongwartz: This conditional statement is the same as just writing ` if ( !oldData )`. Here's why.

Imagine `oldData` is falsey, and `newData` Is also falsey. This statement then says:
```
if ( true && false || true && true )
```.
But if `oldData ` is falsey, and `newData` is truthy, it says:
```
if ( true && true || false && true )
```

So the overall value becomes true if `oldData` is false, regardless of the value of `newData`.


One more point - in this case you're okay, I think, but multiple conditional evaluations can get ambiguous. For example:
```
if ( var1 && var2 || var3 )
```
This can be interpreted as `(var1 && var2) || var3`, but also as `var1 && (var2 || var3)`. When using multiple conditional checks, it's a best practice to use brackets around the checks you actually want to make, to avoid such ambiguity.

PR #83, opened 2018-01-24
homework week2 

malnajar:


week2/index.js
```diff
@@ -0,0 +1,163 @@
...
+                    // No such file or directory
+                    console.log("Sorry, this file doesn't exist.");
+                } else if (err.code === "EACCES") {
+                    // the user don't have a permission
+                    console.log("Sorry, but you don't have permission.");
+                } else {
+                    // any other Error
+                    console.log(` unknown error occurred: ${err.message}`);
+                }
+                process.exit();
```
jasongwartz: You could also reject the promise here, and let the caller decide what to do - exit, or try and handle it gracefully some other way (like creating a missing file).

week2/index.js
```diff
@@ -0,0 +1,163 @@
+"use-strict";
+
+const fs = require('fs');
+const TODO_FILE = process.env['TODO_file'] || 'todo.json'
+const HELP_MENU = 'help.txt';
```
jasongwartz: This variable name is a bit weird, since it's not the help menu - it's just the filename of the help file. I would have gone with `HELP_FILENAME`, or something similar. Not `HELP_FILE`, because it's not the contents - just the filename.

jasongwartz: You could also do something like:
```
const FILENAMES = { help: "help.txt", todos: "todo.json" }
```
That way you just have a look-up table of all the filenames you might need, to be able to do `readFile( FILENAMES.help )`.

week2/index.js
```diff
@@ -0,0 +1,163 @@
...
+                    // No such file or directory
+                    console.log("Sorry, this file doesn't exist.");
+                } else if (err.code === "EACCES") {
+                    // the user don't have a permission
+                    console.log("Sorry, but you don't have permission.");
+                } else {
+                    // any other Error
+                    console.log(` unknown error occurred: ${err.message}`);
+                }
+                process.exit();
```
jasongwartz: If we exit the process during a failure here, we might actually lose data (the new todo we were trying to enter), and data loss is just about the worst sin that code can commit.

Like above, i think it might make more sense to reject the promise, and let the caller retry or do something else.

week2/index.js
```diff
@@ -0,0 +1,163 @@
...
+        })
+    })
+}
+
+let addTodoItem = (filename, todoList, itemText) => {
+    todoList.push({
+        TASK_NUMBER: todoList.length + 1,
+        TODO_TASK: itemText
+    });
+    writeFiles(filename, JSON.stringify(myTodoList)).then(() => console.log("\nDone Adding the task on your TODO list.\n\n")).then(() => askingLoop());
```
jasongwartz: This one line is too long, it should be split up like: 
```
writeFiles(filename, JSON.stringify(myTodoList)).then(() => {
  console.log("\nDone Adding the task on your TODO list.\n\n")
}).then(() => askingLoop());
```

Or, to keep all the `.then`s in line:
```
writeFiles(filename, JSON.stringify(myTodoList))
  .then(() => console.log("\nDone Adding the task on your TODO list.\n\n"))
  .then(() => askingLoop());
```

week2/index.js
```diff
@@ -0,0 +1,163 @@
...
+                process.exit();
+            }
+            resolve();
+        })
+    })
+}
+
+let addTodoItem = (filename, todoList, itemText) => {
+    todoList.push({
+        TASK_NUMBER: todoList.length + 1,
```
jasongwartz: Why `TASK_NUMBER` and `TODO_TASK`? Surely it would make sense to either have `TODO_NUMBER` and `TODO_TASK`, or `TASK_NUMBER` and `TASK_TODO` or something? The names are inconsistent.

jasongwartz: Also, I think it's a mistake to assign a `TASK_NUMBER` here. Arrays are ordered lists, so the order is always maintained - that means you can calculate the task number based on the position in the array, instead of having to manage the task numbers yourself (like you do with the for loop on line 61).

I think what you're hinting at is unique IDs, which will get covered in the databases class, so don't worry about it for now.

week2/index.js
```diff
@@ -0,0 +1,163 @@
...
+
+function askingLoop(data) {
+
+    rl.question('Write your commend > ', function (data) {
+
+        switch (data.toString().trim()) {
+            case 'help':
+                readFiles(HELP_MENU).then((data) => {
+                    console.log(data);
+                    process.stdout.write('\nWrite your commend > ')
```
jasongwartz: You use `process.stdout.write` here, but `readline` elsewhere - you should pick one style, and always do it in the same way.

week2/index.js
```diff
@@ -0,0 +1,163 @@
...
+                console.log(`\n----------------------------------------------------------------------------------\n`);
+                myTodoList.map(task => { console.log(`    ${task.TASK_NUMBER}:${task.TODO_TASK}`) });
+                console.log(`\n----------------------------------------------------------------------------------\n`);
+                process.stdout.write('\nWrite your commend > ');
+                break;
+            case 'add':
+
+                rl.question('\nWrite a new task or "E" to exit> ', function (task) {
+
+                    if (task.toString().trim() != "E" && task.toString().trim() != "e") {
```
jasongwartz: You could put this check at the beginning of the function, instead of having to check inside every block.

week2/index.js
```diff
@@ -0,0 +1,163 @@
...
+                break;
+            case 'list':
+                console.log(`\n----------------------------------------------------------------------------------\n`);
+                myTodoList.map(task => { console.log(`    ${task.TASK_NUMBER}:${task.TODO_TASK}`) });
+                console.log(`\n----------------------------------------------------------------------------------\n`);
+                process.stdout.write('\nWrite your commend > ');
+                break;
+            case 'add':
+
+                rl.question('\nWrite a new task or "E" to exit> ', function (task) {
```
jasongwartz: What if my new task is called `"E"`? Seems like an unnecessary limitation, since I could use Ctrl-C to force cancel if I made a mistake.

jasongwartz:
Can you please push new changes for this weeks' homework into this PR, so we can see what you've changed since the comments?
Creating a new PR means we can't see the diff.

jasongwartz:
I can't run it:

```
readline.js:186
    input.on('data', ondata);
          ^

TypeError: Cannot read property 'on' of undefined
    at new Interface (readline.js:186:11)
    at Object.createInterface (readline.js:65:10)
    at Object.<anonymous> (/Users/jasongwartz/test.js:89:21)
    at Module._compile (module.js:660:30)
    at Object.Module._extensions..js (module.js:671:10)
    at Module.load (module.js:573:32)
    at tryModuleLoad (module.js:513:12)
    at Function.Module._load (module.js:505:3)
    at Function.Module.runMain (module.js:701:10)
    at startup (bootstrap_node.js:194:16)
```

jasongwartz:
Looks like you're not catching promise rejections somewhere:

```
Sorry, this file doesn't exist, but new file has been created : todo.json
(node:37826) UnhandledPromiseRejectionWarning: undefined
(node:37826) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:37826) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

week2/index.js
```diff
@@ -0,0 +1,157 @@
...
+
+const fs = require('fs');
+const FILENAMES = { help: "help.txt", todo: "todo.json" }
+
+let readFiles = (filename) => {
+    return new Promise((resolve, reject) => {
+        fs.readFile(filename, "utf8", (err, data) => {
+            if (err) {
+                if (err.code === "ENOENT") {
+                    console.log(`Sorry, this file doesn't exist, but new file has been created : ${filename}`);
```
jasongwartz: To clarify my comment earlier - READING from a non-existent file doesn't create that file. But WRITING to a non-existent file will create a new file.

week2/index.js
```diff
@@ -0,0 +1,157 @@
...
+            default:
+                console.log('Sorry, Wrong input kindly use "help" command for more information');
+
+        }
+        askingLoop(data)
+    });
+
+}
+
+readFiles(FILENAMES.todo).then((data) => {
```
jasongwartz: You have no `catch()` for the promise rejection here, which is why it's throwing the error I posted in a comment.

week2/index.js
```diff
@@ -0,0 +1,157 @@
...
+const FILENAMES = { help: "help.txt", todo: "todo.json" }
+
+let readFiles = (filename) => {
+    return new Promise((resolve, reject) => {
+        fs.readFile(filename, "utf8", (err, data) => {
+            if (err) {
+                if (err.code === "ENOENT") {
+                    console.log(`Sorry, this file doesn't exist, but new file has been created : ${filename}`);
+                    let fileContent = [];
+                    let filePath = `${filename}.json`;
```
jasongwartz: Since `filename` is `todos.json`, the file you're reading + writing from is called `files.json.json`.

week2/index.js
jasongwartz: Also, you shouldn't write a log message that says "something HAS been done" before you've actually done it, because it's not true yet. On this line, you could say "Sorry, this file doesn't exist, but I will attempt to create it: ", and later on, after it's been done successfully, you can say that it "has been created".

PR #88, opened 2018-02-26
Jamshid's week1-hw I think done

jamsrahimi:
Hi George,
I think I completed the homework.
Let me know if I messed up somewhere.

### DIFF
--- a/week1/homework/src/index.js
+++ b/week1/homework/src/index.js
@@ -1,3 +1,73 @@
 'use strict';
 
-// Write the homework code in this file
+const http = require('http');
+
+let state = 10;
+const PORT = 8080;
+const server = http.createServer(handleRequest);
+
+server.listen(PORT, () => {
+  console.log(`Server started at: ${stateUrl}`);
+});
+
+function handleRequest(request, response) {
+  console.log('on request', request.url);
+  switch (request.url) {
+    case '/':
+      response.setHeader('Content-Type', 'text/html');
+      response.write('<h1>Main Page</h1>');
+      break;
+    case '/state':
+      response.setHeader('Content-Type', 'text/html');
+      response.write(`<h1>Current state of the application is: ${state}</h1>`);
+      break;
+    case '/add':
+      response.setHeader('Content-Type', 'text/html');
+      response.write(`<h1>1 is added to the current state: ${state + Number(1)}</h1>`);
+      state = state + 1;
+      break;
+    case '/subtract':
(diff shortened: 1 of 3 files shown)


week1/homework/src/index.js
```diff
@@ -1,3 +1,73 @@
...
+      response.write('<h1>Main Page</h1>');
+      break;
+    case '/state':
+      response.setHeader('Content-Type', 'text/html');
+      response.write(`<h1>Current state of the application is: ${state}</h1>`);
+      break;
+    case '/add':
+      response.setHeader('Content-Type', 'text/html');
+      response.write(`<h1>1 is added to the current state: ${state + Number(1)}</h1>`);
+      state = state + 1;
```
GeorgeSapkin: You can also change the `state` first and then write it to response so as not to duplicate the logic.

Also you don't need to wrap `1` in a `Number`.

week1/homework/src/index.js
```diff
@@ -1,3 +1,73 @@
...
+      break;
+    case '/subtract':
+      response.setHeader('Content-Type', 'text/html');
+      response.write(`<h1>1 is subtracted from the current state: ${state - '1'}</h1>`);
+      state = state - 1;
```
GeorgeSapkin: Same as above. And when adding numbers there's no need to wrap `1` in a string.

week1/homework/src/index.js
```diff
@@ -1,3 +1,73 @@
...
+      break;
+    case '/reset':
+      response.setHeader('Content-Type', 'text/html');
+      state = 10;
+      response.write(`<h1>Current state of the application is reseted: ${state}</h1>`);
+      break;
+    default:
+      response.setHeader('Content-Type', 'text/html');
+      response.write('<h1>Error 404 "Page not found"</h1>');
```
GeorgeSapkin: You forgot to set `response.statusCode` to 404.

week1/homework/src/index.js
```diff
@@ -1,3 +1,73 @@
...
+ * Response: "OK" in HTML format
+ * This should set the state back to '10'
+ */
+// const resetUrl = 'http://localhost:8080/reset';
+
+/* Any other URL
+ * Response: return error code 404: 'Not found' with a friendly message and do
+ * not change the state variable
+ */
+// const badUrl = 'http://localhost:8080/bad';
```
GeorgeSapkin: There's no need to include the whole assignment in code.

PR #89, opened 2018-02-27
Node week1

kenan0x0:
Hopefully didn't mess things up :)

### DIFF
--- a/week1/homework/src/index.js
+++ b/week1/homework/src/index.js
@@ -2,7 +2,7 @@
 
 "use strict";
 
-const state = 10;
+let state = 10;
 
 const http = require("http");
 
@@ -18,19 +18,16 @@ function handleRequest(request, response) {
             response.write(`<h1>This is the original state: ${state}</h1>`);
             break;
         case addUrl:
-            let add = state + 1;
-            console.log(add);
+            state++;
             response.write(`<h1>OK</h1>`);
             break;
         case subtractUrl:
-            let subtract = state - 1;
-            console.log(subtract);
+            state--;
             response.write(`<h1>OK</h1>`);
             break;
         case resetUrl:
-            let reset = state;
-            console.log(reset);
-            response.write(`<h1>Back to the original state: ${reset}</h1>`);
+            state = 10;
+            response.write(`<h1>Back to the original state: ${state}</h1>`);
             break;
         default:
             response.statusCode = 404;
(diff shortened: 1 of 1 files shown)


week1/homework/src/index.js
```diff
@@ -1,3 +1,53 @@
-'use strict';
+/* eslint-disable no-unused-vars */
```
GeorgeSapkin: Why? There aren't any unused variables.

week1/homework/src/index.js
```diff
@@ -1,3 +1,53 @@
...
 
-// Write the homework code in this file
+"use strict";
+
+const state = 10;
```
GeorgeSapkin: You are making `state` a constant. How are you going to modify it afterwards?

week1/homework/src/index.js
```diff
@@ -1,3 +1,53 @@
...
+
+const http = require("http");
+
+const url = "http://localhost:8080";
+
+function handleRequest(request, response) {
+    response.setHeader("Content-Type", "text/html");
+    switch (url + request.url) {
```
GeorgeSapkin: You can also just check `request.url` directly.

week1/homework/src/index.js
```diff
@@ -1,3 +1,53 @@
...
+        case "http://localhost:8080/":
+            response.write(`<em><h1>This is the main page</h1></em>`);
+            break;
+        case stateUrl:
+            response.write(`<h1>This is the original state: ${state}</h1>`);
+            break;
+        case addUrl:
+            let add = state + 1;
```
GeorgeSapkin: You are not modifying the state. If I call `/add` twice after starting the program I expect `state` to be 12. In your code `state` is _always_ 10.

week1/homework/src/index.js
```diff
@@ -1,3 +1,53 @@
...
+            console.log(add);
+            response.write(`<h1>OK</h1>`);
+            break;
+        case subtractUrl:
+            let subtract = state - 1;
```
GeorgeSapkin: Similar to above.

week1/homework/src/index.js
```diff
@@ -1,3 +1,53 @@
...
+            console.log(subtract);
+            response.write(`<h1>OK</h1>`);
+            break;
+        case resetUrl:
+            let reset = state;
```
GeorgeSapkin: Similar to above.

week1/homework/src/index.js
kenan0x0: I added it so that eslint does not underline the last variable which is badUrl,,, I didn't use it as you see. (But yes I could delete it because I never used it anywhere).

PR #90, opened 2018-02-27
K3M1RD1N's Homework week 1

Kemirdin:
Dear George after Watching your video, I have just tried my homework, I hope this would be correct.

### DIFF
--- a/week1/homework/index.js
+++ b/week1/homework/index.js
@@ -1,5 +1,5 @@
 'use strict';
-let state = 10;
+const state = 10;
 const http = require('http');
 const server = http.createServer();
 
--- a/week1/homework/index.js
+++ b/week1/homework/index.js
@@ -31,7 +31,7 @@ server.on('request', (request, response) => {
       setResponse('Error 404, page not found');
   }
 });
-const port = 8080;
+const port = 3030;
 server.listen(port, () => {
-  console.log('Listening on', port);
+  console.log('Listening on port 3030...', port);
 });
--- a/week1/homework/index.js
+++ b/week1/homework/index.js
@@ -10,7 +10,7 @@ server.on('connection', () => {
 server.on('request', (request, response) => {
   console.log('requesting', request.url);
   function setResponse(header) {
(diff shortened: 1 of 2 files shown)


Kemirdin:
Dear George, Let me know if I made my homework wrong.

week1/homework/index.js
```diff
@@ -0,0 +1,39 @@
...
+      setResponse(state--);
+      break;
+    case '/reset':
+      setResponse((state = 10));
+      break;
+    case '/state':
+      setResponse(state);
+      break;
+    default:
+      setResponse('Error 404, page not found');
```
GeorgeSapkin: You forgot to set `request.statusCode` to 404.

week1/homework/index.js
```diff
@@ -0,0 +1,39 @@
...
+  function setResponse(header) {
+    res.setHeader('Content-Type', 'Text-html');
+    res.write(
+      `<h1 style="background-color:blue;color:white;font-size:46px;">${header}</h1>`
+    );
+    res.end();
+  }
+  switch (req.url) {
+    case '/add':
+      setResponse(state++);
```
GeorgeSapkin: You are using post-increment so this will return `state` first and _then_ increment it. So even though state will end up incremented it will still pass previous state to `setResponse`. That is if initial state is 10, this will pass 10 to `setResponse` even though the new state will be 11. You should have used pre-increment instead: `++state` which will increment `state` first and then pass it to `setResponse`.

week1/homework/index.js
```diff
@@ -0,0 +1,39 @@
...
+      break;
+    case '/remove':
+      setResponse(state--);
```
GeorgeSapkin: Similar to above.

week1/homework/index.js
```diff
@@ -0,0 +1,39 @@
...
+      break;
+    case '/reset':
+      setResponse((state = 10));
```
GeorgeSapkin: Double parentheses `((` and `))` is unnecessary.

PR #91, opened 2018-03-01
Node.js Week1 homework added (EMRE)

arasmr:
I think it`s done !

### DIFF
--- a/week1/homework/src/index.js
+++ b/week1/homework/src/index.js
@@ -1,7 +1,5 @@
 'use strict';
 
-'use strict';
-
 const http = require('http');
 
 // state number
--- a/week1/homework/src/index.js
+++ b/week1/homework/src/index.js
@@ -1,3 +1,62 @@
 'use strict';
 
-// Write the homework code in this file
+'use strict';
+
+const http = require('http');
+
+// state number
+let state = 10;
+
+// add and subtract function
+function add() {
+  state += 1;
+}
+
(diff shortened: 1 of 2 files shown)


week1/homework/src/index.js
```diff
@@ -1,3 +1,60 @@
...
+    case '/reset':
+      response.setHeader('Content-Type', 'text/html');
+      const resetState = 10;
+      state = resetState;
+      response.write('<h1>OK</h1>');
+      response.write(`<h1>${state}</h1>`);
+      break;
+    // for bad url
+    default:
+      if (request.url !== '/' || 'state' || 'reset' || 'add' || 'subtract') {
```
GeorgeSapkin: There is no need for this check since you are already excluding those cases by using the `switch` statement with those cases before.

week1/homework/src/index.js
```diff
@@ -1,3 +1,60 @@
...
+    case '/subtract':
+      response.setHeader('Content-Type', 'text/html');
+      subtract();
+      response.write('<h1>OK</h1>');
+      response.write(`<h1>${state}</h1>`);
+      break;
+    case '/reset':
+      response.setHeader('Content-Type', 'text/html');
+      const resetState = 10;
+      state = resetState;
```
GeorgeSapkin: Or you could just say `state = 10` :stuck_out_tongue_closed_eyes:

PR #104, opened 2018-03-06
Kenan week2 Node

kenan0x0:
PS. This is an unfinished homework still working on the remove command! will update gradually.

### DIFF
--- a/week2/homework/src/index.js
+++ b/week2/homework/src/index.js
@@ -1,3 +1,85 @@
-'use strict';
+"use strict";
 
-// Write the homework code in this file
+const fs = require("fs");
+
+function writeFile(path, text) {
+    return new Promise(function (resolve, reject) {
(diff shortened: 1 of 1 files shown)


week2/homework/src/index.js
```diff
@@ -61,19 +61,44 @@ function help() {
...
+            .then(function (data) {
+                console.log("To Dos:")
+                console.log(data.toString());
+        })
+        break;
+    case "remove":
+        readFile(location)
+            .then(function (data) {
+                let info = data.toString().split("\n");
+                info.splice(arg - 1, 1);
```
GeorgeSapkin: What if `arg` is `-100` or `'cat'`? You are not checking if it's a valid value.

kenan0x0:
Added all required functions (I hope), and handled all edge cases from the arg not being a number to it having a negative value.

PR #92, opened 2018-03-01
week1 homework

fadikurdi:

PR #105, opened 2018-03-08
homework week1 &week2 ( week2 is not finished );

fadikurdi:


GeorgeSapkin:
I've already left feedback for week 1 homework in #92 so I'm not going to repeat it here. In the future make new branch for each homework.

PR #108, opened 2018-03-09
Node.js Week2 Homework added(EMRE)

arasmr:
I think it is OK.

### DIFF
--- a/week2/homework/src/index.js
+++ b/week2/homework/src/index.js
@@ -1,3 +1,129 @@
 'use strict';
 
 // Write the homework code in this file
+
+const fs = require('fs');
+
+const toDoFile = 'toDo.txt';
+
+function readFile() {
+    return new Promise(
+        (resolve, reject) => fs.readFile(toDoFile, (err, data) => {
+            if (err) {
+                reject(err);
+            } else {
+                resolve(data.toString());
+            };
+        })
+    );
+}
+
+function appendFile(value) {
+    return new Promise(
+        (resolve, reject) => fs.appendFile(toDoFile, `${value}\n`, (err, data) => {
+            if (err) {
+                reject(err);
+            } else {
+                resolve(data);
+            };
+        })
+    );
+}
+
+function writeFile(...val) {
+    return new Promise(
+        (resolve, reject) => fs.writeFile(toDoFile, `${val.join('\n')}`, (err, data) => {
+            if (err) {
+                reject(err);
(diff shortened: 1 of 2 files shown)


week2/homework/src/index.js
```diff
@@ -1,3 +1,129 @@
...
+            } else {
+                resolve(data);
+            };
+        })
+    );
+}
+
+function listToDo() {
+    readFile()  
+        .then(data => { data !== '' ? console.log(`\nTo - Dos: \n\n${data}`) : console.log('\nThere is nothing in the To-Do list!\n') })
```
GeorgeSapkin: Try to break-up longer lines into multiple lines for better readability.

week2/homework/src/index.js
```diff
@@ -1,3 +1,129 @@
...
+            };
+        })
+        .catch(console.error);
+}
+
+function removeToDo() {
+    readFile()
+        .then(data => {
+            let myData = data.split('\n');
+            myData.splice(input - 1, 1);
```
GeorgeSapkin: What if I call `remove -100`? Also it's better to pass values as function arguments and not as global variables like `input`.

week2/homework/src/index.js
```diff
@@ -1,3 +1,129 @@
...
+
+function printHelp() {
+    console.log(`
+    HackYourFuture Node.js Week 2 - Lecture To-Do App
+
+    Options:
+
+    add    [to-do]                   add a to-do item
+    help                             show help text
+    remove [line index] + ['to-do']  remove to-do
```
GeorgeSapkin: You probably meant `remove [line index]   remove to-do`.

week2/homework/src/index.js
```diff
@@ -1,3 +1,129 @@
...
+}
+
+function updateToDo() {
+    readFile()
+        .then((data) => {
+            let willUpdate = data.split('\n');
+            if (isNaN(input)) {
+                printHelp();
+            } else {
+                console.log(`\nDone!  "${willUpdate[input - 1]}" is updated as "${update}"\n`);
```
GeorgeSapkin: What if I call `update -100 something`? Also it's better to pass values as function arguments and not as global variables like `input` and `update`.

week2/homework/src/index.js
arasmr: Yeah. I meant remove(line index).

PR #93, opened 2018-03-01
node.js-week1 mahmutKaya

mahmutkaya:

PR #109, opened 2018-03-09
Mahmutkaya

mahmutkaya:


GeorgeSapkin:
I've already left feedback for week 1 homework in #93 so I'm not going to repeat it here. In the future make new branch and a PR for each homework.

week2/homework/index.js
```diff
@@ -0,0 +1,64 @@
...
+            (err, data) => resolve(err ? '' : data.toString())
+        )
+    );
+}
+
+function remove(...i) {
+    readFile()
+        .then(content => {
+             content.splice(i, i+1);
+        writeFile(content)
```
GeorgeSapkin: `writeFile` returns a promise that should return inside then.

Also try indent your code correctly.

week2/homework/index.js
```diff
@@ -0,0 +1,64 @@
...
+function readFile() {
+    return new Promise(
+        resolve => fs.readFile(
+            fileName,
+            (err, data) => resolve(err ? '' : data.toString())
+        )
+    );
+}
+
+function remove(...i) {
```
GeorgeSapkin: `remove` has only 1 argument. Why is are you using the [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) here?

week2/homework/index.js
```diff
@@ -0,0 +1,64 @@
...
+      .then(data => console.log(`to-doS:\n ${data}`))
+      .catch(console.error);
+    break;
+  case "list":
+    readFile(fileName).then(content => {
+      console.log("toDoS:\n", content.toString());
+    });
+    break;
+  case "remove":
+    remove(args - 1);
```
GeorgeSapkin: What if `args` is going to be an array? Let's say `[1, 2, 3]`.

week2/homework/index.js
```diff
@@ -0,0 +1,64 @@
...
+    break;
+  case "reset":
+        fs.writeFile(fileName, '', console.log('all toDo items are removed!'));
```
GeorgeSapkin: Or you could just delete the file.

PR #94, opened 2018-03-02
Elmira's week1 nodejs HW

elmira202003:

PR #112, opened 2018-03-09
Elmira

elmira202003:
Dear George, I apologize for delay, I tried to do this hw with json file, and I have some question. tomorrow I'll work more on it and improve my code. thank you for your attention.


GeorgeSapkin:
I've already left feedback for week 1 homework in #94 so I'm not going to repeat it here. In the future make a new branch and a PR for each homework.

week2/homework/index.js
```diff
@@ -0,0 +1,93 @@
...
+        })
+    })
+}
+
+function remove() {
+    readFile()
+        .then((data) => {
+            let arr = data.split('\n')
+            let item = arr.length - 1;
+            arr.splice(item, 1);
```
GeorgeSapkin: `remove` should remove the item specified by the user not the last item.

week2/homework/index.js
```diff
@@ -0,0 +1,107 @@
...
+            if (err)
+                reject(err);
+            else
+                resolve(data);
+        })
+    })
+}
+
+function remove() {
+    if (!args) {
```
GeorgeSapkin: Both `if` and `else` cases do exactly the same thing.

PR #103, opened 2018-03-05
Homework2  node js hamza zydan

zydan1:
hello  george can you please accept my pull request

PR #118, opened 2018-03-14
homework3 node.js

zydan1:
hello george i really tried to write the code better but i encounter  a lot of problems like i couldn't understand the (await) in the lecture code and how node work with classes , i will do my best to understand it during the weekend  , and i will write the homework again .


week3/homework/src/index.js
```diff
@@ -1,3 +1,78 @@
 'use strict';
+let express = require('express');
+let app = express();
+let uuid = require('uuid/v4')
+app.use(express.json());
+let { readFile, writeFile } = require('./todo.js');
+
+app.get('/READ', (req, res) => {
```
GeorgeSapkin: Why is this `/READ` instead of `/todos`?

Also it should return read to-dos to the client through response `res` and not just write them to the console.

week3/homework/src/todos.json
```diff
@@ -0,0 +1,25 @@
...
+        "done" : false
+        
+    },
+    {
+        "name": "shopping",
+        "id": "4343-3523",
+        "done" : false
+        
+    }
+]
```
GeorgeSapkin: This file shouldn't be committed since it may contain user-sensitive data. I've already described this in #103

week3/homework/todos.json
```diff
@@ -0,0 +1 @@
+[]
```
GeorgeSapkin: Same as above.

week3/homework/src/index.js
```diff
@@ -1,3 +1,78 @@
...
+        console.log(JSON.parse(data))
+    })
+})
+app.delete("/todos/:id/done", (req, res) => {
+    let id = req.params.id;
+
+
+    readFile('./todos.json').then((data) => {
+        let newData = JSON.parse(data);
+        for (let i = 0; i < newData.length; i++) {
```
GeorgeSapkin: Why not use `find` instead?

week3/homework/src/index.js
```diff
@@ -1,3 +1,78 @@
...
+
+            console.log(newData[i].name)
+
+        }
+
+    })
+});
+app.delete('/todos', (req, res) => {
+    writeFile('./todos.json', []).then((data) => {
```
GeorgeSapkin: Again, you're not writing anything into response.

week3/homework/src/index.js
```diff
@@ -1,3 +1,78 @@
...
+})
+app.post("/todos/:id/done", (req, res) => {
+    let id = req.params.id;
+
+
+    readFile('./todos.json').then((data) => {
+        let newData = JSON.parse(data);
+        for (let i = 0; i < newData.length; i++) {
+            if (newData[i].id == id) {
+                newData[i].done === true;
```
GeorgeSapkin: You are not assigning `true` here. You are comparing `done` to it.

week3/homework/src/index.js
```diff
@@ -1,3 +1,78 @@
...
+})
+app.delete("/todos/:id/done", (req, res) => {
+    let id = req.params.id;
+
+
+    readFile('./todos.json').then((data) => {
+        let newData = JSON.parse(data);
+        for (let i = 0; i < newData.length; i++) {
+            if (newData[i].id == id) {
+                newData[i].done === false;
```
GeorgeSapkin: Same as above, you are not assigning, you are comparing here.

PR #121, opened 2018-03-17
not finished

mahmutkaya:
I would like to commit after completed but unfortunately I couldn't finish before homework deadline.

### DIFF
--- a/week3/homework/src/index.js
+++ b/week3/homework/src/index.js
@@ -1,3 +1,111 @@
-'use strict';
+"use strict";
 
-// TODO: Write the homework code in this file
+const express = require("express");
+const uuidv4 = require('uuid/v4');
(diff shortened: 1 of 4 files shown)


GeorgeSapkin:
You can still finish it.

week3/homework/src/index.js
```diff
@@ -1,3 +1,103 @@
...
+    res.json(data);
+    res.end();
+  });
+});
+
+app.get("/todos/:id", (req, res) => {
+  const getTodoId = req.params.id;
+  readFile()
+    .then(todos => {
+      const getTodo = todos.filter(todo => todo.id === getTodoId);
```
GeorgeSapkin: Or you could use `find` since you only want one item.

PR #123, opened 2018-03-19
Diana homework Node .js week3

DianaAyoub:
sorry for being late ... there is still a problem in my understanding of branches..

[index.js.txt](https://github.com/HackYourFuture/Node.js/files/1827380/index.js.txt)


GeorgeSapkin:
I think you should really try to figure out git flow with branches and PRs. This PR is very confusing right now. Maybe ask some of your classmates for help?

GeorgeSapkin:
The code doesn't make any sense. `Todo` class in the lecture has none of the methods used. Also we were not using `mongoose` yet. Is this copy-pasted from somewhere?

PR #129, opened 2018-04-10
update main readme

mkruijt:


remarcmij:
@GeorgeSapkin:  Good point, I found plain text versions here:

https://creativecommons.org/2011/04/15/plaintext-versions-of-creative-commons-licenses-and-cc0/

PR #132, opened 2018-04-16
homework week1

gisaac85:
All done!
============================
 √ /state returns 10
  √ /add returns 11
  √ /subtract returns 9
  √ /reset returns 10
  √ querying undefined URL returns 404 Not Found
  √ /add, /reset returns 10
  √ /subtract, /reset returns 10 (101ms)
  √ /add, /add, /add, /subtract returns 12 (128ms)
  √ /subtract, /subtract, /reset, /add, /subtract, /add returns 11 (150ms)
  √ /add, /add, /add, /add, /add, /add, /add, /add, /add, /add returns 20 (176ms)
  √ /subtract, /subtract, /subtract, /subtract, /subtract, /subtract, /subtract, /subtract, /subtract, /subtract returns 0 (179ms)

  11 tests passed

### DIFF
--- a/week1/homework/src/server.js
+++ b/week1/homework/src/server.js
@@ -10,14 +10,13 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
-
+    response.setHeader('Content-Type', 'application/json');
     switch (request.url) {
       case '/state':
         state = 10;
         let number = JSON.stringify({
           state: state
         });
-        response.setHeader('Content-Type', 'application/json');
         response.write(`${number}`);
         break;
 
@@ -25,7 +24,6 @@ function createServer(port) {
         let add = JSON.stringify({
           state: ++state
         });
-        response.setHeader('Content-Type', 'application/json');
         response.write(`${add}`);
         break;
 
@@ -33,7 +31,6 @@ function createServer(port) {
         let subtract = JSON.stringify({
           state: --state
         });
-        response.setHeader('Content-Type', 'application/json');
         response.write(`${subtract}`);
         break;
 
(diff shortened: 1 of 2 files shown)


week1/homework/src/server.js
```diff
@@ -10,11 +10,56 @@ function createServer(port) {
...
     // TODO: Write your homework code here
-  });
 
+    switch (request.url) {
+      case '/state':
+        state = 10;
+        let number = JSON.stringify({
+          state: state
+        });
+        response.setHeader('Content-Type', 'application/json');
```
GeorgeSapkin: You can also put the `setHeader` outside of the switch so you don't have to repeat it.

week1/homework/src/server.js
```diff
@@ -10,11 +10,50 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
-  });
+    response.setHeader('Content-Type', 'application/json');
+    switch (request.url) {
+      case '/state':
+        state = 10;
```
alxbrg: You're re-assigning `10` to `state` here. Which means that this route actually does the same thing as the `/reset` one.

alxbrg: You're reassigning `10` to `state` here. That means that this route does the same thing as the `/reset` route, i.e. resetting `state` to `10`.

week1/homework/src/server.js
```diff
@@ -10,11 +10,50 @@ function createServer(port) {
...
+        let number = JSON.stringify({
+          state: state
+        });
+        response.write(`${number}`);
```
alxbrg: You don't need to use a template string. `JSON.stringify` already return a `string`.

week1/homework/src/server.js
GeorgeSapkin: Overlooked that one.

PR #134, opened 2018-04-17
Homework Week 1

WaelAlnadeem:


week1/homework/src/server.js
```diff
@@ -1,20 +1,47 @@
...
- * will fail.
- */
 function createServer(port) {
   let state = 10;
-
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    response.setHeader('Content-Type', 'application/json');
+    if (request.url === '/state') {
+      state = 10;
```
alxbrg: You're re-assigning 10 to state here. Which means that this route actually does the same thing as the /reset one.

week1/homework/src/server.js
```diff
@@ -1,20 +1,47 @@
...
-/* `createServer` MUST return an instance of `http.Server` otherwise the tests
- * will fail.
- */
 function createServer(port) {
   let state = 10;
-
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    response.setHeader('Content-Type', 'application/json');
+    if (request.url === '/state') {
```
alxbrg: `else if` works perfectly well, but for brevity and readability, we might want to prefer `switch` cases where we have one variable and many possible values.

week1/homework/src/server.js
WaelAlnadeem: yeah if you wrote state it will return 10 and if you wrote reset it will return 10 , what's wrong with it ?

week1/homework/src/server.js
WaelAlnadeem: ok , I'll change it to switch :)

week1/homework/src/server.js
alxbrg: You already declared `state` and assigned it the value `10` at line 5. You don't need to reassign `10` here, but more importantly, every time I connect to the `/state` route, `state` will be reassigned the value `10`. 

You can test it: start your server, go to `/add`, now `state` will be 11, then go to `/state`. The `/state` route should only give you the value of `state` without modifying it. But you'll see that it will tell you that `state` is equal to `10`, while you were expecting `11`.

Does it make sense?

WaelAlnadeem: yes , got it

PR #136, opened 2018-04-19
homework create main server

zaherqasqas:
I'm still working on it ......


alxbrg:
No problem. Let me know if you're stuck.

alxbrg:
Looks very good to me. I really like that you used a function, good job!

Now, just because you seem to get the hang of it: notice that `sendJsonData` is called with exactly the same arguments 4 times. The only time it's called with different ones is at line 37 `sendJsonData(404, { error: 'Not found' })`. 

We could make the code even cleaner and not repeat ourselves at all in several fashions, for instance by simply having a `sendState` function which would be equal to `sendJsonData(200, { state: state })`, or by separating the concerns: (1) updating the state and (2) sending the response. It would look like this, assuming we have defined a function called `isValidUrl` that checks if the url is in our list of valid urls:

```js
// 1. update the state
switch (request.url) {
    case '/add':
        state++;
        break;
    case '/subtract':
        state--;
        break;
    case '/reset':
        state = 10;
        break;
}

// 2. send the response
if (isValidUrl(request.url)) {
    sendJsonData(200, { state: state });
} else {
    sendJsonData(404, { error: 'Not found' });
}
```
Of course that implies checking the url twice, so maybe you'll disagree, but I thought you might find that solution interesting since you're being smart about your code.

PR #138, opened 2018-04-20
Update server.js

Aya-Alabrash:


week1/homework/src/server.js
```diff
@@ -10,6 +10,36 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    console.log(request.url);
+    function switchStates(x) {
+      response.setHeader(`Content-Type`, `application/json`);
```
GeorgeSapkin: Since content type is always `application/json` you could write this line only once outside of `switchState`.

PR #139, opened 2018-04-20
Update server.js

hadeel43:


week1/homework/src/server.js
```diff
@@ -9,9 +9,35 @@ function createServer(port) {
   let state = 10;
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    let States = (state) => {
+      response.setHeader('Content-Type', 'application/json');
```
GeorgeSapkin: Since content type is always `application/json` you could write this line only once outside of `switchState`.

week1/homework/src/server.js
```diff
@@ -9,9 +9,35 @@ function createServer(port) {
   let state = 10;
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    let States = (state) => {
```
GeorgeSapkin: Function names in JS are usually lower case.

PR #152, opened 2018-04-26
Node Week 2 HW

zaherqasqas:


week2/homework/src/commander.js
```diff
@@ -0,0 +1,80 @@
...
+                console.info('TODOS is EMPTY NOW!!!');
+        } else {
+            console.log('Item number is out of range ');
+        }
+    })
+program
+    .command('remove <item>')
+    .action(async function (value) {
+        const data = await readFileWithPromise(TODO_FILE, 'utf8').catch(() => '[]');
+        let todos = JSON.parse(data);
```
alxbrg: You could put those 2 lines in a function since you're using them for almost every command.

PR #154, opened 2018-04-26
muratWeek2Hw

MURATKAYMAZ56:
nodejs week2 hw needs to check out.

### DIFF
--- a/week2/homework/src/handler.js
+++ b/week2/homework/src/handler.js
@@ -1,0 +1,63 @@
+const fs = require('fs');
+const fileName = './to-dos.json';
+
+function add(todo){
+
+    var json = read();
+    json.push(todo);
+
+    save(json);
+}
+
+function read(){
+
+    if(fs.existsSync(fileName)){
+        return require(fileName);
+    } else {
+        return [];
+    }
+}
+
+function remove(index){
+    
+    var json = read();
+
+    if(index >= 0 && index < json.length){
+
+        json.splice(index, 1);
+
+        save(json);
+    }
+}
+
+function reset(){
+    fs.unlinkSync(fileName);
+}
+
+function save(json){
+
+    fs.writeFile(fileName, JSON.stringify(json), function (err) {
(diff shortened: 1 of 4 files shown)


week2/homework/src/index.js
```diff
@@ -1,3 +1,58 @@
...
+    .action(function (param) {
+        handler.remove(param);
+        console.log("removed by index: " + param);
+    });
+
+program
+    .command('list')
+    .description('list all to-dos')
+    .action(function () {
+        console.log(handler.read());
```
alxbrg: It's hard to see what you're logging here, because you're logging a function call. I would do: 
```js 
const todos = handler.read();
console.log(todos);
```

week2/homework/src/handler.js
```diff
@@ -0,0 +1,63 @@
...
+    if(index >= 0 && index < json.length){
+
+        json.splice(index, 1);
+
+        save(json);
+    }
+}
+
+function reset(){
+    fs.unlinkSync(fileName);
```
alxbrg: Maybe handle the error with the callback here?

week2/homework/src/index.js
```diff
@@ -1,3 +1,58 @@
...
+    .action(function (param) {
+        program.help();
+    });
+
+program
+    .command('add [todo]')
+    .description('add to-do')
+    .action(function (param) {
+        handler.add(param);
+        console.log('\"' + param + '\" saved');
```
alxbrg: One problem I see here is that the `add` method is `async` (not as such but it calls `save` which calls the async `writeFile` method) and you're not waiting for its resolution to log `'\"' + param + '\" saved'`. So if there's an error, the user will first see the 'saved' message and then the error. 

You could make the method actually `async` and `await` it, or pass it a callback and do your `console.log` inside of it. I would use the second option, so that your handler methods only care about returning data or an error, and then the `action` functions would take care of logging both your success messages and your errors. 

Does that make sense?

PR #155, opened 2018-04-27
Sarbast hw week2 + bonuses except Commander

SarKurd:


SarKurd:
I'm sure i created a new branch from Master then pushed, i'll make another one now

PR #158, opened 2018-04-28
Week 2

EmadMRI:


week2/homework/src/index.js
```diff
@@ -1,3 +1,73 @@
...
+const readFileWithPromise = promisify(readFile);
+const writeFileWithPromise = promisify(writeFile);
+
+const [ , , cmd, ...args] = process.argv;
+
+const TODO_FILE = 'todos.json'; 
+
+async function main(){
+  const data =JSON.parse( await readFileWithPromise(TODO_FILE, 'utf8').catch(() => '[]'));
+  const elementKey = parseInt(args)-1;
```
alxbrg: As it turns out, `parseInt` will work if you give it an array (taking the first element of that array as its argument), but it would make more sense to do `parseInt(args[0]) - 1`, as it is more inline with what we're doing and with the signature of this function which is supposed to parse strings, not arrays.

week2/homework/src/index.js
```diff
@@ -1,3 +1,73 @@
...
+
+    case 'reset':
+      await writeFileWithPromise(TODO_FILE, JSON.stringify([]));
+    break; 
+
+    case 'update':
+  
+      if(data.length >= elementKey)
+      {
+        args[0] = '';
```
alxbrg: This is a bit odd. You have other (cleaner) ways of selecting the `args` that you actually want here... `Array.prototype.shift` for instance...

PR #159, opened 2018-04-28
Sarbast HW week2 plus bonuses except Commander

SarKurd:


alxbrg:
What happened there? I don't see your changes...

PR #161, opened 2018-04-28
Jason homework week2

thwjason:
Hello - my apologies for the late pull request; I forgot to do this even though I completed it earlier in the day.

The code copied on slack doesn't work. After some trying, I have made it work.

However, I then tried to implement the remove/update keys. I know for certain you have to use the slice/splice methods, but implementation seems a bit spotty. Will need help to sort it. J.


week2/homework/src/index.js
```diff
@@ -1,3 +1,68 @@
...
+      const removeTodo = args.join(' ');
+      todos.slice(removeTodo);
+      await writeFileWithPromise(TODO_FILE, (JSON.stringify(todos)));
+      break;
+    }
+    case 'update': {
+      const data = await readFileWithPromise(TODO_FILE, 'utf8').catch(() => '[]');
+      const todos = JSON.parse(data);
+      const updatedTodo = args.join(' ');
+      todos.splice(updatedTodo);
```
alxbrg: Kind of the same problem as above, but also: `splice` should have 3 arguments here (you should read the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice)): 
1. 1st argument is the index of the array where you want your new element to be; 
2. 2nd argument is the number of elements you want to remove from the array; 
3. 3rd argument is the value you want to insert. 

So we want our element to be inserted at the provided `index`. That will be our first argument. Then we want to remove it, so the 2nd argument will be `1`. And then we'll want to insert the value that's in the args. 

I've already given you a lot of info, let me know if it is still unclear for you.

thwjason:
Many thanks for your comments.

I have done the corrections and they now work - unsure however if they have been pushed here. Please do let me know if it hasn't and then I will sort it out.

thwjason:
Thanks, but couldn't have done it without yourself, Samir and George.

PR #162, opened 2018-04-29
Sobhi Week2 HM

Sob7i:


week2/homework/src/index.js
```diff
@@ -1,3 +1,68 @@
...
+const writeFileWithPromise = promisify(writeFile);
+const TODO_FILE = 'todo.json';
+async function main() {
+    const [, , cmd, ...args] = process.argv;
+    const removed = args[0];
+    switch (cmd) {
+        case 'add':
+            {
+                const data = await readFileWithPromise(TODO_FILE, 'utf8').catch(() => '[]');
+                const todos = JSON.parse(data);
```
alxbrg: Like I said on Slack, those two lines should be in a function since they are used multiple times. You could call this function `getTodos` or `readTodos` for instance.

PR #163, opened 2018-04-29
ISAAC week3 HW

gisaac85:


week3/homework/package.json
```diff
@@ -13,5 +13,11 @@
     "eslint-plugin-node": "^6.0.0",
     "eslint-plugin-promise": "^3.5.0",
     "eslint-plugin-standard": "^3.0.1"
+  },
+  "dependencies": {
+    "express": "^4.16.3",
+    "fs": "0.0.1-security",
```
GeorgeSapkin: `fs` is a built-in module you don't need to install it using `npm`.

week2/homework/src/index.js
```diff
@@ -1,3 +1,3 @@
 'use strict';
 
-// TODO: Write the homework code in this file
+// TODO: Write the homework code in this file
```
GeorgeSapkin: Why did you change this file from week 2 here?

GeorgeSapkin:
You can push changes to the same branch and they will appear in this PR.

gisaac85:
But is my PR is in good acceptable state??

week3/homework/src/index.js
```diff
@@ -1,3 +1,192 @@
...
+//READ ALL TODOs
+app.get('/todos', async (req, res) => {
+  const todos = await readTodos();
+  res.json(todos);
+  res.end();
+});
+
+//READ TODO BY ID
+app.get('/todos/:id', async (req, res) => {
+  try {
```
alxbrg: These `try` and `catch` blocks are a bit weird to me. They are meant to catch run-time errors. We should not get any of those here if our code checks out.

week3/homework/src/index.js
```diff
@@ -1,3 +1,192 @@
...
+    const id = req.params.id;
+    const todos = await readTodos();
+    const todo = todos.find(x => x.id === id);
+    if (typeof todo === 'undefined') {
+      res.json({
+        'Error Message': 'NO such ID !!'
+      });
+    }
+    res.json(todo);
```
alxbrg: Since you already returned a response above if the `todo` was not found, you should not call `res.json` again. This part of the code should be unreachable if we don't need/want it to be executed. What you could do, is `return` on line 64. Does that make sense?

week3/homework/src/index.js
```diff
@@ -1,3 +1,192 @@
...
+    res.json({
+      'Error Name': err.name,
+      'Error Message': err.message
+    });
+  }
+
+});
+
+//DELETE TODO BY ID
+app.delete('/todos/:id', async (req, res) => {
```
alxbrg: Comments above also apply here.

week3/homework/src/index.js
```diff
@@ -1,3 +1,192 @@
...
+  }
+
+});
+
+//markAsNotDone TODO
+app.delete('/todos/:id/done', async (req, res) => {
+  try {
+    const id = req.params.id;
+    const todos = await readTodos();
+    const newTodos = todos.find(todos => todos.id === id);
```
alxbrg: Besides the changes you need to make, I see the same lines of code repeating. Maybe you could have some functions to help you do these actions on the todos. For instance a `findTodo` function which would accept the `request` as a param and return either throw an error (which you could `catch` 😉 ) or the `todo` that was found.

week3/homework/src/index.js
gisaac85: done!

week3/homework/src/index.js
```diff
@@ -1,3 +1,192 @@
...
+//CREATE TODO
+app.post('/todos', async (req, res) => {
+  const newTodo = req.body;
+  newTodo.id = uuid();
+  newTodo.done = false;
+  const todos = await readTodos();
+  todos.push(newTodo);
+  await writeTodos(todos);
+  res.json(todos);
+  res.end();
```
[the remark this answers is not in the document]
gisaac85: done!

week3/homework/src/index.js
gisaac85: can i use res.send on line 64 when ID is not exist !!?

week3/homework/src/index.js
```diff
@@ -1,3 +1,107 @@
...
+
+//DELETE TODO BY ID
+app.delete('/todos/:id', async (req, res) => {
+
+  const [, , ,
+    todo,
+    todos
+  ] = await findTodo(req);
+  if (typeof todo === 'undefined') {
+    return new Error(res.send('NO SUCH ID'));
```
alxbrg: See previous comment.

week3/homework/src/index.js
```diff
@@ -1,3 +1,107 @@
...
+    id,
+    bodyReq,
+    done,
+    todo,
+    todos
+  ] = await findTodo(req);
+  const newTodo = bodyReq;
+  newTodo.id = id;
+  if (typeof todo === 'undefined') {
+    return new Error(res.send('NO SUCH ID'));
```
alxbrg: See previous comment.

week3/homework/src/functions.js
```diff
@@ -0,0 +1,75 @@
...
+  return writeFile(TODO_FILE, JSON.stringify(todos));
+}
+
+//Find Todo
+async function findTodo(req) {
+  const id = req.params.id;
+  const todos = await readTodos();
+  const todo = todos.find(todos => todos.id === id);
+  const done = req.body.done;
+  return await [
```
alxbrg: `return await` does not really make sense. You should simple `return`. You will have to `await` the resolution of the function when you call it.

week3/homework/src/functions.js
```diff
@@ -0,0 +1,75 @@
...
+function writeTodos(todos) {
+  return writeFile(TODO_FILE, JSON.stringify(todos));
+}
+
+//Find Todo
+async function findTodo(req) {
+  const id = req.params.id;
+  const todos = await readTodos();
+  const todo = todos.find(todos => todos.id === id);
+  const done = req.body.done;
```
alxbrg: Not sure what's the point of this. Can you explain?

week3/homework/src/functions.js
```diff
@@ -0,0 +1,75 @@
...
+
+//MarkTodo Function
+async function markTodo(req, res, doneFlag) {
+  const [, ,
+    done,
+    todo,
+    todos
+  ] = await findTodo(req);
+  if (typeof todo === 'undefined') {
+    return new Error(res.send('NO SUCH ID'));
```
alxbrg: Returning a `new Error` here does not make sense to me. Simply return `res.send('NO SUCH ID')`. Makes sense? Also: you should return a json for the errors as well, like so: `{ error: 'NO SUCH ID' }`. It will help us use that error when using this in the browser when we code our front-end application.

week3/homework/src/index.js
```diff
@@ -1,3 +1,107 @@
...
+  res.json(todos);
+
+});
+
+//READ TODO BY ID
+app.get('/todos/:id', async (req, res) => {
+
+  const [, , , todo, ] = await findTodo(req);
+  if (typeof todo === 'undefined') {
+    return new Error(res.send('NO SUCH ID'));
```
alxbrg: See previous comment.

week3/homework/src/functions.js
```diff
@@ -0,0 +1,75 @@
...
+    done,
+    todo,
+    todos
+  ];
+}
+
+//MarkTodo Function
+async function markTodo(req, res, doneFlag) {
+  const [, ,
+    done,
```
alxbrg: You're not using this variable anywhere.

week3/homework/src/functions.js
```diff
@@ -0,0 +1,75 @@
...
+  const todos = await readTodos();
+  const todo = todos.find(todos => todos.id === id);
+  const done = req.body.done;
+  return await [
+    id,
+    req.body,
+    done,
+    todo,
+    todos
+  ];
```
alxbrg: Not sure why you return so many variables 😊 I would have just returned `todo` and `todos`.

week3/homework/src/functions.js
gisaac85: when i try to delete ( , , done , ) i get error !!

week3/homework/src/functions.js
gisaac85: because i need them in markTodo and updatTodo functions !!!

gisaac85:
can you please see my last changing !!

week3/homework/src/functions.js
alxbrg: Your `findTodo` should only return what you need and are going to use. So `findTodo` should only return a `todo` and the `todos` (I don't think you are using `id`, `req.body` and `done` anywhere).

gisaac85:
Hi @alxbrg  
thanks for your constructive feedback..
Yes I see that returning [todo,todos] from function findTodo() are enough to do all tasks 💯 
I hope now you accept my PR 👍

PR #165, opened 2018-05-01
my week 3 homework in Node.js

Aya-Alabrash:


week3/homework/src/index.js
```diff
@@ -1,3 +1,96 @@
...
+  await writeTodos(todos);
+  res.json(todos);
+});
+
+//update a todo
+app.put('/todos/:id', async (req, res) => {
+  const id = req.params.id;
+  const newTodo = req.body;
+  const todos = await readTodos();
+  todos[id].description = newTodo.description;
```
GeorgeSapkin: Great idea using IDs as keys. If you're curious about a more efficient way to do that, have a look at [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

Also you're not validating the input anywhere. What if the ID doesn't exist? In case of updating, you're going to create a new record even though you should have returned an error. Same thing for other routes that have an ID. Have a look at [app.param](https://expressjs.com/en/4x/api.html#app.param) for a way of validating individual params only once.

Aya-Alabrash: Thank you I'll work more on it 👍

week3/homework/src/index.js
```diff
@@ -1,3 +1,109 @@
...
+  const todos = await readTodos();
+  res.json(todos);
+});
+
+//Read a todo by ID
+app.get('/todos/:id', async (req, res) => {
+  const id = req.params.id;
+  const todos = await readTodos();
+  if (id in todos) { res.json(todos[id]); }
+  else { res.json(`this ID isn't exist `); }
```
GeorgeSapkin: When something is not found, you can also set the response status code to 404 to help clients easily figure out what went wrong.

Also, since we're expecting a JSON object back, you could wrap the errors into something like:

```json
{
  "error": "To-do not found"
}
```

Aya-Alabrash: yes i think it's better too 
i'll see it & thank you for advice 💕
if there is anything too i'll be glad to learn more

PR #167, opened 2018-05-02
Week 3 homework 

hadeel43:


week3/homework/src/index.js
```diff
@@ -1,3 +1,119 @@
...
+  res.json({});
+});
+
+//update a todo
+app.put('/todos/:id', async (req, res) => {
+  const id = req.params.id;
+  const todos = await readTodos();
+  const updatedTodo = req.body;
+  if (todos.hasOwnProperty(id)) {
+    todos[id].description = updatedTodo.description;
```
alxbrg: @remarcmij Pointed to me that the way you are storing the todos doesn't actually fit with the way you are updating them. You are getting something like this from the requests:
```
{
  "todo": {
    "description": "(todo description)"
  }
}
```
Since you store the body of the request directly, your `JSON` file looks like this after adding a todo:
```
{
    "4cdcf95a-37e9-4d0f-aa06-4ed9f5818b57": {
        "todo": {
            "description": "(todo description)"
        },
        "done": false
    }
} 
```
Then if you update it like you're doing it now, it looks like this: 
```
{
    "4cdcf95a-37e9-4d0f-aa06-4ed9f5818b57": {
        "todo": {
            "description": "(todo description)"
        },
        "description": "updated"
        "done": false
    }
} 
```
I missed it at first, sorry about that :S

week3/homework/src/index.js
```diff
@@ -1,3 +1,119 @@
...
+    await writeTodos(todos);
+    res.json(todos[id]);
+  }
+  else {
+    res.json(`id not found`);
```
alxbrg: Also, it's best if you return a JSON encoded object for your errors, so that once you get the error message, your front-end application can easily parse and use it. But you'll learn more about that later on. For now, simply return something like this: 
```
{
  "error": "id not found"
}
```

week3/homework/src/index.js
hadeel43: I do every update to each todos[id] so, it will make the update of description on todos[id].description only and I try it on my lap and show me another output I can send you a screenshot of it

remarcmij: Hi Hadeel, if I may chip in. You are right that your application works fine when you post TODOs like you do. However, the assignment gave a specific format for the request body format:

```json
{
  "todo": {
    "description": "(todo description)"
  }
}
```

See this link:

https://github.com/HackYourFuture/Node.js/tree/master/week3#request-body-format

Therefore your implementation does not comply with the specification. This will give issues if people attempt to use your backend based on the specification.

PR #174, opened 2018-05-04
homework2 is renewed

Semih21:


week2/homework/src/index.js
```diff
@@ -1,3 +1,70 @@
...
 
 // TODO: Write the homework code in this file
+
+const fs = require('fs');
+
+let fileName = 'toDo.json';
+// console.log(fileName);
+function readFile() {
+    return new Promise(
+        resolve => fs.readFile(
```
alxbrg: You should provide the encoding (`'utf8'`) here. That way you don't have to do `data.toString()` below.

week2/homework/src/index.js
```diff
@@ -1,3 +1,70 @@
...
+        )
+    );
+}
+
+function remove() {
+    readFile()
+        .then(content => {
+            let newArray = content.split('\n');
+            let i = newArray.length-1
+            newArray.splice(i, 1);
```
alxbrg: You are always removing the last element of the array. At line 60 you call the `remove` function with `removeArgs` but it is not used here.

week2/homework/src/index.js
```diff
@@ -1,3 +1,70 @@
...
+           let new2 =  newArray.join('\n')    
+            writeFile(new2);
+            console.log(content);
+        })
+}
+
+function writeFile(...text) {
+    return new Promise(
+        (resolve, reject) => fs.appendFile(
```
alxbrg: You are assuming that the file exists here. What happens if you try to `appendFile` and there is no `toDo.json`?

week2/homework/src/index.js
```diff
@@ -1,3 +1,70 @@
...
+    });
+    break;
+  case "remove":
+    remove(removeArgs);
+    break;
+  case "reset":
+        fs.writeFile(fileName, '', console.log('all toDo items are removed!'));
+        break;
+  default: "help"
+    fs.readFile("help.txt", "utf8", function(err, data) {
```
alxbrg: I don't see the `help.txt` file in this PR?

week2/homework/src/index.js
```diff
@@ -1,3 +1,70 @@
...
+    );
+}
+
+const cmd = process.argv[2];
+const args = process.argv.slice(3);
+const removeArgs = args[0];
+
+switch (cmd) {
+  case "add":
+    writeFile(...args)
```
alxbrg: By using the rest operator `...` here and in your function definition, you're essentially just passing the `args` array. You could simply do `writeFile(args)` and then define your `writeFile` function like so: `function writeFile(text)`. Does that make sense?

alxbrg:
My bad, I seem to understand you're not in class 14. Disregard my comment regarding getting help. We did something a bit different with class 14, that's why I thought you did it on your own! 😊

PR #177, opened 2018-05-05
Murat node week3 hw

MURATKAYMAZ56:
murat nodejs week3 hw.
late is better than nothing:)


alxbrg:
One thing though: your PR contains the homework for week2. You should branch out of master each time you make a PR and only include the homework for the relevant week, otherwise it's a bit of a pain to review.

PR #178, opened 2018-05-05
05 05 week3 emad

EmadMRI:


week3/homework/src/index.js
```diff
@@ -1,3 +1,73 @@
...
+  await writeTodos(todos);
+  response.json(todos);
+});
+
+// Get todo by ID  'readTodo'
+app.get('/todos/:id', async(request, response) => {
+  const todos = await readTodos();
+  const id = request.params.id;
+  await writeTodos(todos);
+ id in todos ? response.json(todos[id]) : response.json('This todo does not exist!') ;
```
alxbrg: You should return an object like so: `{ error:  'This todo does not exist!' }`. This will become important when you'll start looking into front-end stuff for instance.

week3/homework/src/index.js
```diff
@@ -1,3 +1,73 @@
...
+  todos[uuid()] = newTodo;
+  await writeTodos(todos);
+  response.json(todos);
+});
+
+// Get todo by ID  'readTodo'
+app.get('/todos/:id', async(request, response) => {
+  const todos = await readTodos();
+  const id = request.params.id;
+  await writeTodos(todos);
```
alxbrg: Why are you writing todos here?

week3/homework/src/index.js
```diff
@@ -1,3 +1,73 @@
...
+ id in todos ? response.json(todos[id]) : response.json('This todo does not exist!') ;
+});
+
+//  'markAsDone '
+app.post('/todos/:id/done', async(request, response) => {
+  const todos = await readTodos();
+  const id = request.params.id;
+  todos[id].done = true;
```
alxbrg: What happens if the `id` is not in `todos`?  You somewhat handle this case two lines below but do modify the `todos` object before doing so...

PR #179, opened 2018-05-06
initial commit

thwjason:


week3/homework/src/index.js
```diff
@@ -1,3 +1,102 @@
...
+  const todos = await readTodos();
+  res.json(todos);
+});
+
+// Read a todo by ID
+app.get('/todos/:id', async (req, res) => {
+  const id = req.params.id;
+  const todos = await readTodos();
+  if (id in todos) {
+    res.json(todos[id]);
```
alxbrg: By default, you are using an array to store your todos (see line 25). `uuid` returns a `string` so your ids are going to look like `f64f2940`. This is not really how arrays are supposed to be used. Instead you should an object as your default todo store so that each property of that object will be an id. Does that make sense?

thwjason:
I didn't see the add mark as done/not done functionality - please see if the one I added works.

Also changed line 25 to sort out storing to object, but unsure if this is the way to do it...

PR #180, opened 2018-05-06
Week3

Rushkurdi:


week3/homework/src/index.js
```diff
@@ -1,3 +1,128 @@
...
+  const todos = await readTodos();
+  res.status(200).json(todos);
+});
+
+// Read a todo by ID
+app.get('/todos/:id', async (req, res) => {
+  const id = req.params.id;
+  const todos = await readTodos();
+  const todo = todos.find(x => x.id === id);
+  res.json(todo);
```
alxbrg: What happens when you cannot find the `id`? 😛

week3/homework/src/index.js
```diff
@@ -1,3 +1,128 @@
...
+    });
+  } else {
+    todos[updateIndex].description = req.body.description;
+    await writeTodos(todos);
+    res.status(200).json(todos);
+
+  }
+});
+// markAsDone
+app.post('/todos/:id/done', async (req, res) => {
```
alxbrg: This route is very similar to the the 'markAsNotDone' route. The only difference is that you're doing `todos[updateIndex].done = true` here and `todos[updateIndex].done = false` there. Maybe you could have a function with an argument `done` which will be true or false. That way you don't have to copy-paste almost the same code in two places. Does that make sense?

PR #183, opened 2018-06-27
Week1 homework

PKnar:


week1/homework/src/server.js
```diff
@@ -8,8 +8,38 @@ const http = require('http');
 function createServer(port) {
   let state = 10;
 
+  let sendResponse = (response) => {
```
alxbrg: When you're not reassigning your variable, you should use `const`. Also: you should define that function outside of the `createServer` one, otherwise, `sendResponse` gets defined every time we call `createServer()`.

PKnar: Got it and fixed it, thanks for the feedback @alxbrg  :+1:

PR #184, opened 2018-06-28
Added homework week1

neveenatik:


week1/homework/src/server.js
```diff
@@ -10,6 +10,30 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    function sendResponse(status, stateValue, contentType) {
```
alxbrg: Try not to define functions inside of other functions. `sendResponse` here gets defined every time we call `http.createServer`.

week1/homework/src/server.js
```diff
@@ -10,6 +10,30 @@ function createServer(port) {
...
+      state = stateValue;
+      response.writeHead(status, { 'Content-Type': contentType });
+      response.write(`{ "state": ${state}}`);
```
alxbrg: This will work but if your object gets more complicated, this way of doing things can get hairy. We have a nice util to translate an object into a JSON string: `JSON.stringify()`.

week1/homework/src/server.js
neveenatik: Thanks I corrected that. and this is to make performance better and not keep the stack occupied. is this right?

alxbrg: Yes, it is about performance. Defining a function like this is the same as assigning any kind of value to a variable. If you do it inside a function, the JS engine will have to recreate that function (or object, array, etc.) each time it is called. Without going into the details, you pay that with memory and CPU usage. If you're interested, this is a good article: https://www.reaktor.com/blog/javascript-performance-fundamentals-make-bluebird-fast/

PR #185, opened 2018-06-28
added homework week1

Wael-Alhomsi:


week1/homework/src/server.js
```diff
@@ -10,6 +10,32 @@ function createServer(port) {
...
+      case '/reset':
+        state = 10;
+        serverResponse();
+        break;
+      default:
+        response.writeHead(404, { 'Content-Type': 'application/json' });
+        response.end(JSON.stringify({ error: 'Not found' }));
+        break;
+    }
+    function serverResponse() {
```
alxbrg: You shouldn't define functions inside of other functions. `serverResponse` here gets defined every time we call `http.createServer`.

PR #186, opened 2018-06-28
added homework week1

Dumie1:
I still have 3 tests failing out of 11. Still yet to fix that.

### DIFF
--- a/week1/homework/src/server.js
+++ b/week1/homework/src/server.js
@@ -8,8 +8,35 @@ const http = require('http');
 function createServer(port) {
   let state = 10;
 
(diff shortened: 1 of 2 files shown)


week1/homework/src/server.js
```diff
@@ -10,6 +10,36 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+
+    function calculate(response) {
```
alxbrg: Try not to define functions inside of other functions. `calculate` here gets defined every time we call `http.createServer`.

PR #187, opened 2018-06-28
Homework week1

ThreePalmTrees:


week1/homework/src/server.js
```diff
@@ -1,17 +1,36 @@
-'use strict';
+"use strict";
```
alxbrg: Why did all of the single quotes got replaced by double quotes? 😛 It may seem like a non-important detail, but you need to respect the style of the code that you are contributing to. When you will work within a team, one style will be enforced and you need to comply with it, otherwise it's a mess!

week1/homework/src/server.js
```diff
@@ -1,17 +1,36 @@
...
-const http = require('http');
+const http = require("http");
 
-/* `createServer` MUST return an instance of `http.Server` otherwise the tests
- * will fail.
- */
 function createServer(port) {
   let state = 10;
-
+  function okResponse(response) {
```
alxbrg: Try not to define functions inside of other functions. `okResponse` here gets defined every time `createServer` is called...

week1/homework/src/server.js
ThreePalmTrees: Thanks for the feedback Alex, 
This is happening because of the ESlint settings, I'll search how to change it to single quotations

PR #188, opened 2018-06-28
adding node-hw-w1

amudar:


week1/homework/src/server.js
```diff
@@ -2,14 +2,38 @@
...
 
-/* `createServer` MUST return an instance of `http.Server` otherwise the tests
- * will fail.
- */
 function createServer(port) {
   let state = 10;
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    const sendJsonData = (statusCode, jsonMessage) => {
```
alxbrg: Since you're always passing the same value (`{ state: state }`), `jsonMessage` could be defined in this function, instead of being passed as an argument. 

Also: you should not define functions inside of other functions. `sendJsonData` here gets defined every time `createServer` is called.

PR #196, opened 2018-07-05
Homework week2 added

amirno1:
Something went wrong with the indetation in VScode, i tried to fix it with the help of Alex but it is'nt yet. So he is going to take a look at it on Sunday. Thanks for your feedback in advance

### DIFF
--- a/week2/homework/src/index.js
+++ b/week2/homework/src/index.js
@@ -1,3 +1,78 @@
 'use strict';
 
-// TODO: Write the homework code in this file
+const { readFile, writeFile } = require('fs');
+
+const { promisify } = require('util');
+
+const readFileWithPromise = promisify(readFile);
+const writeFileWithPromise = promisify(writeFile);
+
+const args = process.argv.slice(2);
+
+const cmd = args[0];
+
+let TODO_PATH = 'todo-json';
+
+function readToDos() {
+	return readFileWithPromise(TODO_PATH, 'utf8')
+		.then(JSON.parse)
+		.catch(() => ([]));
+}
+function writeToDos(toDos) {
+	return writeFileWithPromise(TODO_PATH, JSON.stringify(toDos, null, 2));
+}
+
+switch (cmd) {
+	case 'add':
+		const text = args[1];
+		readToDos()
+			.then(toDos => {
+				toDos.push({ toDo: text });
+				return toDos;
+			})
+			.then(writeToDos);
+		break;
+	case 'remove':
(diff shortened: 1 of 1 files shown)


week2/homework/src/index.js
```diff
@@ -1,3 +1,78 @@
...
+				if (start >= 0 && typeof newText === 'string') {
+			toDos.splice(start, 1);
+			toDos[start] = { toDos: newText };
+		}else {
+			console.info('Please check again the index you passed and the new To Do you want to pass in');
+		}
+		return toDos;
+}).then(writeToDos);
+break;
+	case 'help':
```
alxbrg: Since the `'help'` case is the same as the default, you don't need to specify it. In any case, if you're going to reuse a string several times, you must assign it to a variable. That way you don't have to copy-paste it. Remember: be DRY. If 2 lines of code look exactly the same, you're doing something wrong.

week2/homework/src/index.js
```diff
@@ -1,3 +1,78 @@
...
+				toDos.push({ toDo: text });
+				return toDos;
+			})
+			.then(writeToDos);
+		break;
+	case 'remove':
+		let index = args[1];
+		readToDos()
+			.then(toDos => {
+				if (index >= 0) {
```
alxbrg: What if the index is greater than the number of todos in the list?

week2/homework/src/index.js
```diff
@@ -1,3 +1,78 @@
...
+		break;
+	case 'update':
+		const start = args[1];
+		const newText = args[2];
+		readToDos()
+			.then(toDos => {
+				if (start >= 0 && typeof newText === 'string') {
+			toDos.splice(start, 1);
+			toDos[start] = { toDos: newText };
+		}else {
```
alxbrg: The indentation here is very confusing. I know you are having problems with your code editor, so maybe this is the reason. We'll have a look at it together tomorrow, but the `else` needs to have the same indentation as the `if`.

week2/homework/src/index.js
```diff
@@ -1,3 +1,78 @@
...
+				return toDos;
+			}).then(writeToDos);
+		break;
+	case 'update':
+		const start = args[1];
+		const newText = args[2];
+		readToDos()
+			.then(toDos => {
+				if (start >= 0 && typeof newText === 'string') {
+			toDos.splice(start, 1);
```
alxbrg: Indentation! See my comment below.

PR #198, opened 2018-07-05
Added homework week2

neveenatik:
Thank you for your feedback. I appreciate that.

### DIFF
--- a/week2/homework/src/index.js
+++ b/week2/homework/src/index.js
@@ -23,7 +23,6 @@ async function app() {
 
     case 'add':
       const task = args[1];
-      // let toDos = await readToDos();
       toDos.push({
         task,
         done: false
@@ -47,8 +46,7 @@ async function app() {
       break;
 
     case 'reset':
-      resetToDos()
-        .catch(() => ([]));
+      resetToDos();
       break;
 
     case 'done':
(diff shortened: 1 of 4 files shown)


week2/homework/src/index.js
```diff
@@ -1,3 +1,95 @@
...
+async function app() {
+  const toDos = await readToDos();
+  switch (cmd) {
+    case 'list':
+      readToDos().then(console.log);
+      break;
+
+    case 'add':
+      const task = args[1];
+      // let toDos = await readToDos();
```
alxbrg: You left one comment behind there 😛 .

week2/homework/src/index.js
```diff
@@ -1,3 +1,95 @@
...
+        console.log(remainingToDos);
+      }
+      else {
+        console.log(`Not a valid entry, to check the list type 'list'`);
+      }
+      break;
+
+    case 'reset':
+      resetToDos()
+        .catch(() => ([]));
```
alxbrg: This `catch` is a bit confusing. You are not doing anything with the array that you are returning. If you want to ignore the errors that might be thrown by you `resetToDos` function, return `undefined` or `null` instead of an array, and put a comment to explain what you're doing.

PR #200, opened 2018-07-06
add homework week2

AllouhM:


AllouhM:
Dear George, Thank you very much, it is my bad. I deleted it.

week2/homework/src/index.js
```diff
@@ -1,3 +1,79 @@
...
+
+
+async function content() {
+    const [, , command, ...args] = process.argv;
+    const entry = args.join(' ');
+    switch (command) {
+        case ('add'): {
+            const currentList = await toRead();
+            currentList.push(entry);
+            if (!entry) { console.log('Please add todo after add command'); return args; }
```
alxbrg: Ok but it makes more sense to do that check at the very beginning of your `'add'` case, since we don't need to read the list or to push the entry to the array if it is missing.

week2/homework/src/index.js
```diff
@@ -1,3 +1,79 @@
...
+        case ('reset'): {
+            await toWrite('[]');
+            const data = await toRead();
+            console.log('To do list is empty : ' + data);
+            break;
+        }
+        case ('remove'): {
+            const currentList = await toRead();
+            const remove = new Number(entry) - 1;
+            if (!entry) { return console.log('Please add the number of todo needs to be removed') }
```
alxbrg: Same thing here: you start processing the argument before checking if you even should! 😛

week2/homework/src/index.js
```diff
@@ -1,3 +1,79 @@
...
+            currentList.splice(remove, 1);
+            await toWrite(currentList);
+            console.log(`activity # ${entry} has been removed from list `);
+            break;
+        }
+        case ('update'): {
+            const currentList = await toRead();
+            const toUpdate = new Number(args[0]) - 1;
+            currentList.splice(toUpdate, 1, args[1]);
+            if (!args[1]) { return console.log('Please add a new todo after the number of replaced activity'); }
```
alxbrg: Same comment as above.

week2/homework/src/index.js
```diff
@@ -1,3 +1,79 @@
...
+            await toWrite(currentList);
+            console.log(`activity # ${args[0]} has been updated`);
+            break;
+
+        }
+        case ('Help'):
```
alxbrg: By convention, we use lower cases for option and command names.

week2/homework/src/index.js
```diff
@@ -1,3 +1,79 @@
...
+            break;
+        }
+
+        case ('list'): {
+            const data = await toRead();
+            console.log('Your to do list is : ' + data);
+            break;
+        }
+        case ('reset'): {
+            await toWrite('[]');
```
alxbrg: You already `stringify` your input in the `toWrite` function. You should pass an empty array here, not a string.

AllouhM:
Thank you very much for your useful and informative feedback

PR #201, opened 2018-07-06
added homework week2

zmagzoub:
I fixed the problem. Now only week2 homework is added. Also, There is something missed in my homework in the first pull request, but I've fixed the problem here.


week2/homework/src/index.js
```diff
@@ -1,3 +1,76 @@
...
+      break;
+
+    case 'help':
+      console.log('commands :');
+      console.log('');
+      console.log('add : to add to-do');
+      console.log('remove : to remove a to-do');
+      console.log('reset : to remove all the to-dos');
+      console.log('list : to see all the to do');
+      break;
```
GeorgeSapkin: Why did you duplicate this code? At the very least you could have abstracted it into a `help` function. But better yet you can combine `switch` cases:

```js
switch (cmd) {
  case 'add':
    // do something
    break;
  case 'help':
  default:
    // do something
    break;
}
```

week2/homework/src/index.js
```diff
@@ -1,3 +1,76 @@
...
+        done: false
+      });
+      await writeTodos(todos);
+      const newTodos = await readTodos();
+      console.log(newTodos);
+      break;
+
+    case 'remove':
+      const remove = await readTodos();
+      remove.splice(args[1] - 1, 1);
```
GeorgeSapkin: What if I call `remove` with invalid index like `-1000`, `cat` or even `0`? You're not handling those cases.

PR #202, opened 2018-07-07
Homework week2

ThreePalmTrees:


week2/homework/src/index.js
```diff
@@ -1,3 +1,183 @@
...
+
+  function updateTodo() {
+    allTodos.forEach((todo, index) => {
+      if (index === userIndexInput) {
+        const oldTodo = todo.task;
+        console.log(
+          `updated '${chalk.red(oldTodo)}' with '${chalk.cyan(args[2])}'`
+        );
+        allTodos[index].task = args[2];
+        allTodos[index].done = 'not done';
```
ThreePalmTrees: maybe if I wasn't so blind I could've used **todo.task**, and **todo.done** instead of **allTodos[index].task** and **allTodos[index].done** (lines 147 & 148)

week2/homework/src/index.js
```diff
@@ -1,3 +1,183 @@
...
+      align: 'left'
+    }
+  )
+};
+
+async function main() {
+  const userIndexInput = parseInt(args[1]);
+  const allTodos = await readJSON();
+
+  function addTodo() {
```
GeorgeSapkin: You could have put all these functions outside of `main`. This is not helping with readability.

week2/homework/src/index.js
```diff
@@ -1,3 +1,183 @@
...
+    const userTodoInput = args[1];
+    if (userTodoInput === undefined) {
+      console.log(
+        chalk.red('Error while adding a todo, see this:\n'),
+        helpMessages.add
+      );
+    } else {
+      allTodos.push({
+        task: userTodoInput,
+        done: 'not done'
```
GeorgeSapkin: Why didn't you just use `true` or `false`?

week2/homework/src/index.js
```diff
@@ -1,3 +1,183 @@
...
+      });
+      storeTodos(allTodos);
+      console.log(`added todo: '${chalk.cyan(args[1])}'`);
+    }
+  }
+
+  function removeTodo() {
```
GeorgeSapkin: Doesn't work according to the assignment. The index should be 1-based and not zero-based. When I call it with `remove 1` it should remove the first to-do and not the second one.

week2/homework/src/index.js
```diff
@@ -1,3 +1,183 @@
...
+    if (userIndexInput >= allTodos.length) {
+      console.log(`${chalk.red("didn't find a todo with that index")}`);
+      console.log(helpMessages.remove);
+    }
+    allTodos.forEach((todo, index) => {
```
GeorgeSapkin: You don't need to iterate over all to-dos to delete 1 item.

week2/homework/src/index.js
```diff
@@ -1,3 +1,183 @@
...
+            `${chalk.inverse(`you've already marked '${todo.task}' as done`)}`
+          );
+        } else {
+          todo.done = 'done';
+          console.log(`Successfully marked '${chalk.cyan(todo.task)}' as done`);
+          storeTodos(allTodos);
+        }
+      }
+    });
+    if (userIndexInput >= allTodos.length) {
```
GeorgeSapkin: Why don't you check this in the beginning of the function? And what if the index is less than zero? It that OK?

week2/homework/src/index.js
```diff
@@ -1,3 +1,183 @@
...
+              ? chalk.green.bold(todo.done)
+              : chalk.magenta.bold(todo.done)
+          }`
+        );
+      });
+    }
+  }
+
+  function markTodoAsDone() {
+    allTodos.forEach((todo, index) => {
```
GeorgeSapkin: Why do you iterate over the whole array to update 1 item? The whole point of an array is that you can access item by index. Also, the assignment is to remove and update items using 1-based index and not zero-based.

week2/homework/src/index.js
```diff
@@ -1,3 +1,183 @@
...
+          storeTodos(allTodos);
+        }
+      }
+    });
+    if (userIndexInput >= allTodos.length) {
+      console.log(`${chalk.red("didn't find a todo with that index")}`);
+    }
+  }
+
+  function updateTodo() {
```
GeorgeSapkin: Same issues as with `markTodoAsDone`.

week2/homework/src/index.js
```diff
@@ -1,3 +1,183 @@
...
+        task: userTodoInput,
+        done: 'not done'
+      });
+      storeTodos(allTodos);
+      console.log(`added todo: '${chalk.cyan(args[1])}'`);
+    }
+  }
+
+  function removeTodo() {
+    if (userIndexInput >= allTodos.length) {
```
GeorgeSapkin: What if the index is less than zero? It that OK?

week2/homework/src/index.js
ThreePalmTrees: Thank you for the feedback George, I worked on the homework again and put the functions outside, hope it's more readable now.

week2/homework/src/index.js
ThreePalmTrees: Cause I wanted to show the input as _done_ or _not done_, maybe I could've used true and false then check with an if statement if `true` then log `'done'` or if `false` log `'not done'`.

week2/homework/src/index.js
ThreePalmTrees: Done, fixed all index related issues.

week2/homework/src/index.js
ThreePalmTrees: Removed all the `forEach`'s except for `listTodos()` and modified the code accordingly.

week2/homework/src/index.js
ThreePalmTrees: :heavy_check_mark:

week2/homework/src/index.js
ThreePalmTrees: All fixed too :heavy_check_mark: .

week2/homework/src/index.js
ThreePalmTrees: :heavy_check_mark:

week2/homework/src/index.js
ThreePalmTrees: `const husam = 1;`
node . done 1 :smile:

week2/homework/src/actions/addTodo.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const chalk = require('chalk');
+const helpMessages = require('./helpMessages');
+const storeTodos = require('./storeTodos');
+
+const args = process.argv.slice(2);
```
GeorgeSapkin: You should have passed arguments that each action requires instead of reading a global object.

week2/homework/src/actions/addTodo.js
```diff
@@ -0,0 +1,26 @@
...
+  const userTodoInput = args[1];
+  if (userTodoInput === undefined) {
+    console.log(
+      chalk.red('Error while adding a todo, see this:\n'),
+      helpMessages.add
+    );
+  } else {
+    allTodos.push({
+      task: userTodoInput,
+      done: 'not done'
```
GeorgeSapkin: Why didn't you use a boolean value for `done`?

week2/homework/src/actions/markTodoAsDone.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const chalk = require('chalk');
+const listTodos = require('./listTodos');
+const storeTodos = require('./storeTodos');
+
+function markTodoAsDone(userIndexInput, allTodos) {
+  if (userIndexInput > allTodos.length || userIndexInput < 1) {
```
GeorgeSapkin: Since you're doing this check in several places, maybe you could have placed it in a separate function.

PR #204, opened 2018-07-11
Homework week2

Dumie1:
I have committed only the changes for week2.


week2/homework/src/index.js
```diff
@@ -1,3 +1,82 @@
...
+        .then(JSON.parse)
+        .catch(() => ([]));
+}
+
+function writeTodos(todos) {
+    return writeFileWithPromise(TODO_PATH, JSON.stringify(todos, null, 2));
+}
+
+function myHelper() {
+    console.log('Get help from these commands:1- node.add \'New todo\'   = Add a todo 2- node . delete [Delete item]  = Delete item from the list 3- node . list  = List all items 4- node . update [Item to be updated]  \'New todo\'  = Updating a todo. 5- node . reset  = Reset the list');
```
GeorgeSapkin: You can split a string into multiple lines by using `\n` escape code like this `'line one\nline two\nand so on'`. Alternatively you can use [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) like this:

```js
const multiLineString = `line one
line two`;

File: week2/homework/src/index.js [hunk 2]
```diff
@@ -1,3 +1,82 @@
...
+            .then(readTodos)
+            .then(console.log);
+        break;
+
+    case 'delete':
+        let item = args[1];
+        readTodos()
+            .then(todos => {
+                if (item > 0) {
+                    todos.splice(item, 1);
```
GeorgeSapkin: The assignment was to use a 1-based index to delete and update items so when I call `delete 1` the first item is deleted and not the second one as is with your code. Also the command was supposed to be called `remove`.

Also you're not validating the input what if `item` equals `-100` or `'cat'`?

week2/homework/src/index.js
```diff
@@ -1,3 +1,82 @@
...
+        break;
+
+    case 'list':
+        readTodos().then(console.log);
+        break;
+
+    case 'reset':
+        readTodos()
+            .then(todos => {
+                todos.splice(0, todos.length);
```
GeorgeSapkin: Or you could just write an empty array.

week2/homework/src/index.js
```diff
@@ -1,3 +1,88 @@
...
+    break;
+
+  case 'update':
+    let updateItem = args[1];
+    let newItem = args[2];
+    readTodos()
+      .then(todos => {
+        if (updateItem >= 0 && typeof newItem === 'string') {
+          todos.splice(updateItem, 1);
+          todos[updateItem] = { todo: newItem };
```
GeorgeSapkin: You can use `splice` to replace an item directly.

Dumie1:
@GeorgeSapkin , thank you for the feedback. I have corrected delete and update as required...

Dumie1:
Thank you @GeorgeSapkin , i am sure update now has correct logic.

week2/homework/src/index.js
```diff
@@ -1,3 +1,88 @@
...
+        return todos;
+      }).then(writeTodos);
+    break;
+
+  case 'update':
+    let updateItem = args[1];
+    let newItem = args[2];
+    readTodos()
+      .then(todos => {
+        if (updateItem >= 0 && typeof newItem === 'string') {
```
GeorgeSapkin: What if I try to update an item that has an index that is higher than the length of the todo array? Right now this will insert a new item with that index.

PR #205, opened 2018-07-12
Knar-  homework week 3

PKnar:
For whatever reason, the Postman application does not properly run on my laptop(even after several uninstalls and reinstalls, detecting the problems etc). So I did not get to fully test all the requests, especially app.delete('todos/: id'......) one. Hopefully, those requests make sense and work :thinking:


PKnar:
@GeorgeSapkin I have no idea, to be honest, the only files I worked on are actions.ja and index.js and I worked directly from the src folder.  I see there is only one package-lock.json and 2 packages.json files. Shall I add dependencies to the homework folder's package.json and delete the uploaded package.json one?

GeorgeSapkin:
> Shall I add dependencies to the homework folder's package.json and delete the uploaded package.json one?

Yes.

PKnar:
@GeorgeSapkin  done

PKnar:
@GeorgeSapkin sorry I accidentally closed the pull request. Ok, It's done I removed the file and added the original one.

PKnar:
@GeorgeSapkin  ok I will do it too

PKnar:
@GeorgeSapkin not sure how the package-lock.json got in that folder but I changed it too. I think everything is in order now.

week3/homework/src/index.js
```diff
@@ -1,3 +1,84 @@
...
+    createTodo(todo)
+        .then(data => res.send(data))
+        .catch(err => next(err))
+});
+
+// readList
+app.get('/todos', (req, res) => {
+    readTheList()
+        .then(data => res.send(data))
+        .catch(err => next(err));
```
GeorgeSapkin: `next` is undefined.

week3/homework/src/actions.js
```diff
@@ -0,0 +1,88 @@
...
+const { readFile, writeFile } = require('fs');
+const { promisify } = require('util');
+const uuid = require('uuid/v4');
+const read = promisify(readFile);
+const write = promisify(writeFile);
+const FILE_PATH = 'todo.json';
+const list = {};
+
+
+const fileContent = (todo) => {
```
GeorgeSapkin: `fileContent` is not the best name for a function. Function names should be actions, e.g. `writeToDo`.

week3/homework/src/actions.js
```diff
@@ -0,0 +1,88 @@
...
+        const foundItem = todoList.find(todo => todo[id]);
+        foundItem[id].description = newTodo.description;
+        await fileContent(todoList);
+        return todoList;
+    } catch (error) {
+        throw new Error('Item cannot be found.');
+    }
+};
+
+const markAsDoneOrNotDone = async (id, boolean) => {
```
GeorgeSapkin: `boolean` is a not good name for a variable since it's not really clear what the intention is. Why not `done`?

week3/homework/src/actions.js
```diff
@@ -0,0 +1,88 @@
...
+    const todoList = await readTheList();
+
+    try {
+        const foundItem = todoList.find(todo => todo[id]);
+        foundItem[id].done = boolean;
+        await fileContent(todoList);
+        return todoList;
+    } catch (error) {
```
GeorgeSapkin: This error handling doesn't really distinguish between item not being found and file failing to be written.

week3/homework/src/actions.js
```diff
@@ -0,0 +1,88 @@
...
+        throw new Error('Item cannot be found.')
+    }
+};
+
+const clearTodos = async () => {
+    const todoList = await readTheList();
+
+    if (todoList.length === 0) throw new Error('The list is empty');
```
GeorgeSapkin: What is the purpose of this check?

week3/homework/src/index.js
```diff
@@ -1,3 +1,84 @@
...
+app.put('/todos/:id/done', (req, res, next) => {
+    const { id } = req.params;
+
+    markAsDoneOrNotDone(id, true)
+        .then(data => res.send(data))
+        .catch(err => next(err));
+});
+
+//mark as undone
+app.put('/todos/:id/undone', (req, res, next) => {
```
GeorgeSapkin: According to the assignment this should have been `DELETE /todos/:id/done`. But no big deal.

week3/homework/src/actions.js
```diff
@@ -0,0 +1,88 @@
+'use strict';
+
+const { readFile, writeFile } = require('fs');
+const { promisify } = require('util');
+const uuid = require('uuid/v4');
+const read = promisify(readFile);
+const write = promisify(writeFile);
+const FILE_PATH = 'todo.json';
+const list = {};
```
GeorgeSapkin: Why is this variable defined here?

week3/homework/src/actions.js
```diff
@@ -0,0 +1,88 @@
...
+const fileContent = (todo) => {
+    return write(FILE_PATH, JSON.stringify(todo, null, 2));
+};
+
+const createTodo = async (todo) => {
+    if (!todo.description) throw new Error('Description is required');
+
+    const todoList = await readTheList();
+    const id = uuid();
+    list[id] = todo;
```
GeorgeSapkin: You are adding each new to-do to a global variable and then pushing that object (eventually, with multiple to-dos) to the array. So this method doesn't work as intended.

week3/homework/src/index.js
PKnar: oh right, I added `next` as the third parameter.

week3/homework/src/actions.js
PKnar: Changed this too.

week3/homework/src/actions.js
PKnar: That's right, the thing is at first I used the catch() method in fileContent () function, to display a message if something goes wrong. But the console would not show that error message and would show specific error messages based on the problem. So I thought maybe catch is not necessary. Can you maybe explain that?

week3/homework/src/actions.js
PKnar: oh, it should not be there, I was supposed to pass it in,  delete by id, function like I did in command line app version :sweat_smile:  but I guess I don't need that here because a user would first get the list and then delete based on id. And if the list was empty they would see it, so as I understand this check is not necessary.

week3/homework/src/actions.js
alxbrg: I agree with George here. `boolean` is less explicit than `done`. The point is to be as specific as possible as to *what the variable actually represents*, rather than its type. The type you can document using JSDoc.

week3/homework/src/actions.js
PKnar: 2 reasons :smile:   1. I followed the class example and 2. last week I passed the todo object directly to the list, inside the function, and during the review, Alexandre asked if I can bring it outside the function. So I thought that's what I needed to do here as well  :sweat_smile: especially when in the class example it was outside too. If that's wrong I guess I misunderstood Alexander's comment. Shall I bring it inside the function?

week3/homework/src/actions.js
PKnar: ok I will keep that in mind, I just changed that too

week3/homework/src/actions.js
```diff
@@ -0,0 +1,86 @@
...
+        await writeTodo(todoList);
+        return todoList;
+    } catch (error) {
+        throw new Error('Item cannot be found.')
+    }
+};
+
+const clearTodos = async () => {
+    const todoList = await readTheList();
+    return await writeTodo([]);
```
GeorgeSapkin: `return await` is redundant. You can just `return` instead since return value of `clearTodos` will be a promise anyway.

week3/homework/src/actions.js
GeorgeSapkin: What results do you get when you try to run your own code and add multiple to-dos? Try it and see if you get expected results.

PR #208, opened 2018-07-12
node-w2-hw

amudar:


week2/homework/src/index.js
```diff
@@ -1,3 +1,74 @@
...
+      }
+      else if (value <= 0 || value > todos.length) {
+        console.info(`Enter a number between 1 and ${todos.length}`);
+      }
+      break;
+    }
+
+    case 'help':
+    default:
+      console.info('Here are the options:',
```
GeorgeSapkin: You can split a string into multiple lines by using `\n` escape code like this `'line one\nline two\nand so on'`. Alternatively you can use [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) like this:

```js
const multiLineString = `line one
line two`;

Right now this will print all the help text in a single line.

File: week2/homework/src/index.js [hunk 2]
```diff
@@ -1,3 +1,74 @@
...
+      const data = await readFileWithPromise(toDoFile, 'utf8').catch(() => '[]');
+      const todos = JSON.parse(data);
+      console.info(todos);
+      break;
+    }
+
+    case 'reset': {
+      await writeFileWithPromise(toDoFile, JSON.stringify([]));
+      const data = await readFileWithPromise(toDoFile, 'utf8').catch(() => '[]');
+      const todos = JSON.parse(data);
```
GeorgeSapkin: Why are you reading and parsing the data and then not doing anything with it?

PR #210, opened 2018-07-14
node-w3-hw

amudar:


week3/homework/src/index.js
```diff
@@ -1,3 +1,119 @@
...
+  } else {
+    todos[updateIndex].description = req.body.description;
+    await writeToDos(todos);
+    res.status(200).json(todos);
+
+  }
+});
+
+// markAsDone:
+app.post('/todos/:id/done', async (req, res) => {
```
GeorgeSapkin: The only difference between this handler and the next one is setting `done` to either `true` or `false`. Maybe it could have been a utility function to reduce code duplication?

week3/homework/src/index.js
```diff
@@ -1,3 +1,119 @@
...
+
+// Update a todo by ID:
+app.put('/todos/:id', async (req, res) => {
+  const id = req.params.id;
+  const todos = await readToDos();
+  const updateIndex = todos.findIndex(x => x.id === id);
+  if (updateIndex < 0) {
+    res.status(400).json({ error: 'bad request' });
+  } else {
+    todos[updateIndex].description = req.body.description;
```
GeorgeSapkin: This is not going to work with the [format](https://github.com/HackYourFuture/Node.js/blob/master/week3/README.md#request-body-format) described during the lecture.

PR #212, opened 2018-07-14
Homework week3

Dumie1:
markAsDone and markAsNotDone routes and actions are not fully functioning. However, I will try to fix that.  I also have to handle errors on createTodo function/route.


Dumie1:
I am sorry for that mistake. I have committed my source code without todo.json file.

PR #215, opened 2018-07-16
add homework week3

AllouhM:
hello, this is the complete and updated version of my homework, I am uploading it in new PL as it seems  that the last PL was for a branch that I by mistake created it on brance week2 . Now everything should be in it is right place. thanks

### DIFF
--- a/week3/homework/src/actions.js
+++ b/week3/homework/src/actions.js
@@ -1,0 +1,79 @@
+'use strict';
+const { readFile, writeFile } = require('fs');
+const { promisify } = require('util');
+const uuid = require('uuid/v4');
+const reading = promisify(readFile);
+const writing = promisify(writeFile);
+const TODO_FILE = 'todolist.json';
+
+const readTodos = () => {
+    return reading(TODO_FILE, 'utf8').then(JSON.parse).catch(() => ([]));
+};
+const toWrite = (todo) => {
+    return writing(TODO_FILE, JSON.stringify(todo, null, 2));
+};
+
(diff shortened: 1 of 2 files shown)


week3/homework/src/actions.js
```diff
@@ -0,0 +1,76 @@
...
+const toWrite = (todo) => {
+    return writing(TODO_FILE, JSON.stringify(todo, null, 2));
+};
+
+const createTodo = async (todo) => {
+    if (!todo.description) throw new Error('Please add description to your todo');
+    const todoList = await readTodos();
+    const id = uuid();
+    todo.id = id;
+    todo.status = 'Todo needs action';
```
GeorgeSapkin: Why are you initially writing a string to `status` but then changing it to a boolean? That's a inconsistent.

week3/homework/src/actions.js
```diff
@@ -0,0 +1,76 @@
...
+    currentTodo.description = updateTodo.description;
+    await toWrite(todoList);
+    return todoList;
+};
+const markTodo = async (id, status) => {
+    if (!id) throw new Error('Please enter valid ID');
+    const todoList = await readTodos();
+    const targetTodo = todoList.find(todo => todo.id === id);
+    if (status === false) {
+        return deleteTodo(id);
```
GeorgeSapkin: I would expect marking to-do as undone to not delete it, but just change the status.

AllouhM:
Thank you George for your review, amendments took place as per your review. regards..

PR #216, opened 2018-07-28
Sam node homework

hackoor83:
Hello @alxbrg ,

Thanks for your support.

Here is the latest version, with that only one persisting error. I appreciate your support.

Thank you!


GeorgeSapkin:
@hackoor83 you've pushed homework 2 into the same branch as homework 1. You need to create a new branch from `master` for homework 2.

PR #218, opened 2018-09-04
finished homework node week1

genegr88:
My homework Node.js week-1


week1/homework/src/server.js
```diff
@@ -7,9 +7,42 @@ const http = require('http');
...
+  let body;
+  let statusCode = 200;
+  let message = 'OK';
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    switch (request.url) {
+      case '/state':
+        body = JSON.stringify({state});
+        console.log('route /state', state);
```
J05J4: You should not console.log the result:
https://github.com/HackYourFuture/Node.js/blob/master/week1/homework/README.md#endpoints-to-implement
states you should return the state in JSON format:
```
{
  "state": 10
}
```
This should not be written to the console, but to the response.

J05J4:
Yes, I saw it: i’m improving my response

week1/homework/src/server.js
```diff
@@ -7,9 +7,42 @@ const http = require('http');
...
+        break;
+      case '/reset':
+        state = 10;
+        body = JSON.stringify({state});
+        console.log('route /reset', state);
+        break;
+      default:
+        statusCode = 404;
+        message = 'Not Found';
+        body = JSON.stringify({error: 'Not Found'});
```
J05J4: One test is failing because your message is `'Not *F*ound'` instead of `'Not *f*ound'`

PR #220, opened 2018-09-05
Node.js - Week1 Homework

mehmetscoskun:


week1/homework/src/server.js
```diff
@@ -2,14 +2,47 @@
...
+function createServer() {
   let state = 10;
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    switch (request.url) {
+      case '/state':
+        const stateJson1 = { 'state': state };
+        response.writeHead(200, { 'Content-Type': 'application/json' });
+        response.end(JSON.stringify(stateJson1));
```
J05J4: Why create a `const` variable every time for the response?  

You could also use 

`response. write(JSON.stringify({ 'state': state }));`

We usually use `response.write` to write to the client and use `response.end` to indicate the end.

week1/homework/src/server.js
```diff
@@ -2,14 +2,47 @@
...
+        break;
+
+      case '/add':
+        state++;
+        const stateJson2 = { 'state': state };
+        response.writeHead(200, { 'Content-Type': 'application/json' });
```
J05J4: You are adding a header to every `case` and writing the state as JSON to the response. You could put this in a function, witch you can call from the different `case`s. 

In that way, if you would like to change the key for 'state' in `{'state': state}` into `{'the_state': state}` for example, you have to only do it in one place.

mehmetsaidcoskun:
Hi @J05J4 ,
Thank you for your valuable comments.
I changed my code. I added `response.write` command. Besides I created a function to write contents about `state`. I could not define the `writeContent` function in file level or outside of the `createServer` function. Because if define it in file level, I have to add one more parameter for `response`. That's why I defined it inside `createServer` function.
Thank you :)

PR #222, opened 2018-09-05
week1 homework

sima-milli:


week1/homework/src/server.js
```diff
@@ -9,7 +9,40 @@ function createServer(port) {
   let state = 10;
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    const reqUrl = request.url;
+    if (reqUrl === '/' || reqUrl === '/state') {
+      response.writeHead(200, { 'Content-Type': 'application/json' });
+      const home = {
```
J05J4: You don't have to write the response to a variable, since you only use it once, it makes more cense to write `response.write(JSON.stringify({'state': 'home'}));`

week1/homework/src/server.js
```diff
@@ -9,7 +9,40 @@ function createServer(port) {
...
+        'state': state
+      };
+      response.end(JSON.stringify(home, null, 2));
```
J05J4: You don't need the `null` and `2` parameters to stringify.

You can use `response.end()`, but we usually use `response.write()` to write to the response, and only use `response.end()` to indicate the end of the response.

week1/homework/src/server.js
```diff
@@ -9,7 +9,40 @@ function createServer(port) {
...
+    } else if (reqUrl === '/add') {
+      response.writeHead(200, { 'Content-Type': 'application/json' });
```
J05J4: You are having similar logic in every case of the if statement. You should movr this logic inside of a function and call the function from the different if blocks, passing the right information trough arguments.

week1/homework/src/server.js
```diff
@@ -9,7 +9,40 @@ function createServer(port) {
...
+      const add = {
+        'state': ++state
+      };
+      response.end(JSON.stringify(add, null, 2));
+    } else if (reqUrl === '/subtract') {
```
J05J4: It's OK to use `if` and `else if`, but at some point it could ben more readable (and less code) to use a `switch` with different `case`s.

week1/homework/src/server.js
sima-milli: got it

week1/homework/src/server.js
sima-milli: I'll fix my code, thanks a lot for the review👍

sima-milli: Fixed the code, would you check it after the changes that I made @J05J4

PR #223, opened 2018-09-05
Shadi's homework Week1

shadoo77:


week1/homework/src/server.js
```diff
@@ -1,17 +1,45 @@
...
-
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    let jsonContent = JSON.stringify(state);
+    let msgStatus = 0;
+    switch (request.url) {
+      case '/':
+      case '/reset':
+        state = 10;
+        jsonContent = JSON.stringify({ 'state': state });
```
J05J4: You are creating a response and a status code in every `case` of your `switch`. You should move this duplicate code to a function. In that same function, you could add writing the headers and the writing the state to the response ad well.

PR #225, opened 2018-09-06
BarbarosWeek1

whitepiolin:


week1/homework/src/server.js
```diff
@@ -9,7 +9,31 @@ function createServer(port) {
   let state = 10;
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    function main(statusCode, state) {
+      response.writeHead(statusCode, { 'Content-Type': 'application/json' });
+      response.write(JSON.stringify(state, null, 1));
```
J05J4: You don't really need the `null, 1`, arguments: you could use `JSON.stringify(state)` the other arguments are only for formatting the output, but it makes no difference to the browser who receives the JSON response.

whitepiolin: Thanks for your comments.

PR #226, opened 2018-09-07
Donal's homework1

Donal86:


week1/homework/src/server.js
```diff
@@ -9,7 +9,34 @@ function createServer(port) {
...
+        response.write(JSON.stringify({ state }));
+        break;
+      default:
+        if (request.url === '/') {
+          response.write(JSON.stringify({ state }));
+        }
+        else {
+          response.statusCode = 404;
+          const notFound = { 'error': 'Not found' };
+          response.end(JSON.stringify(notFound, null, 2));
```
J05J4: You don't really need the `null, 2`, arguments: you could use `JSON.stringify(notFound)` the other arguments are only for formatting the output, but it makes no difference to the browser who receives the JSON response.

Donal86: Hi @J05J4 Thanks a lot for your feedback I will follow it for the next homeworks.
 see you!!

PR #228, opened 2018-09-08
week1 homework

acimanx:
ib4: ternaries are evil.

### DIFF
--- a/week1/homework/src/server.js
+++ b/week1/homework/src/server.js
@@ -1,25 +1,33 @@
-'use strict';
+"use strict";
 
-const http = require('http');
+const http = require("http");
 
-const sendState = require('./responses/sendState');
-const sendError = require('./responses/sendError');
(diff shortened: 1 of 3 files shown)


week1/homework/src/server.js
```diff
@@ -2,19 +2,27 @@
...
+const sendError = require('./responses/sendError');
+
+const createServer = port => {
   let state = 10;
 
-  const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+  return http.createServer(({ url }, response) => {
+    url === '/'
+      ? response.writeHead(302, { Location: '/state' })
```
J05J4: This is unreadable with ternarie operators... Please change this to a switch...

acimanx: Hello Josja. 
In my opinion more readable than switch, a bit more readable then if-else. 
No problem I'll fix it today.

PR #229, opened 2018-09-08
Gaffari week1homework

Gaffari02:


week1/homework/src/server.js
```diff
@@ -10,6 +10,32 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    if (request.url === '/state') {
```
J05J4: When you use a lot of if statements it makes more sense to use 
```
switch (request.url) {
    case '/state':
        ...
    case '/add':
        ...
}
```

week1/homework/src/server.js
```diff
@@ -10,6 +10,32 @@ function createServer(port) {
...
+      response.writeHead(200, { 'Content-Type': 'application/json' });
+      const main = { 'state': state };
```
J05J4: You don't need the variable, since you are only using it in `JSON.stringify()`, so you can do `JSON.stringify({'state':state})`

week1/homework/src/server.js
```diff
@@ -10,6 +10,32 @@ function createServer(port) {
...
+      response.end(JSON.stringify(main));
+    }
+    else if (request.url === '/add') {
+      response.writeHead(200, { 'Content-Type': 'application/json' });
```
J05J4: You are writing a head and writing to the response in every case of the if-statement: this is duplicate code; you should use a function for this.

Gaffari02:
Thank you @J05J4 for the feedback. I have added the function and removed duplicate code.

PR #231, opened 2018-09-09
the homework of week1

ozgurleo:


ozgurleo:
@J05J4 
Thanks for taking time to review my codes.

PR #232, opened 2018-09-10
End0620 homework week2

end0620:


week2/homework/src/to-do.json
```diff
@@ -0,0 +1,4 @@
+{
+  "count": 3,
+  "tasks": "1\n22222\n333\n"
```
J05J4: What does this mean? 
If this file isn't available I get an error... Maybe you want to create this file if you need it and it isn't available?

end0620: Im not sure if i can do that in my program because im using asyn functions. so even if i check first if the file exist to create and then write or read from it, is not going to work.

PR #236, opened 2018-09-13
Homework Week2

ozgurleo:


ozgurleo:
Thanks for sharing time @keepitterron 
Actually it works well or I missed your point
I defined variable item as:
let item = process.argv[3];   //which is the definition comes after the comment index to be deleted or element to be added.
and in the remove function, it was used to define to fetch index nummer of element in the todo list and the item is deleted as :
arrayRemove.splice(item,1);
I checked again it works correctly. I define which index i want to delete and it fulfills.
Am i missing a point? Do you mean by "Index is not used in the remove function' index number comes after remove comment isn't it?

keepitterron:
You are absolutely right @ozgurleo !!! Totally my fault 👍 
I've reviewed a bunch of homework today and it was a common mistake, I read `item` in the function and I (wrongly) assumed you did the same.

Apologies. I'll review the code now :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,61 @@
...
+           break;
+};
+function add (item){
+    fs.appendFile('./data.txt', item + '\n', (error)=>{
+        if (error){
+            console.log(error);
+        };
+    });
+};
+function remove (item){
```
keepitterron: I would rename the argument of this function to something else. I was confusing 'cause I expected `item` do be a todo item. Maybe `index` is a better name? :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,61 @@
...
+    let arrayRemove= fs.readFileSync('./data.txt').toString().split('\n');
+    arrayRemove.splice(item,1);
```
keepitterron: Just a note here: you are using javascript's 0-based index.
As a user of your app I don't know that and I expect to provide a 1-based index.

Consider this list:
```
todo 1
todo 2
todo 3
```

`node . remove 1` will remove `todo 2`, do you think this is the expected behavior?

Again, the  function works fine, this is a UX issue :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,61 @@
 'use strict';
 
 // TODO: Write the homework code in this file
+let fs = require('fs');
```
keepitterron: For variables that do not (or should not) change, prefer using `const` over `let`.
Actually, always try to use `cost`, `let` should be used as little as possible.

This is because if you use `const` you cannot redeclare a variable by mistake and break your app without understanding why. If you reuse a variable you declared with `const` the app will break but you will get an error that tells you why, if you re-assign or re-declare a variable with `let`, the compiler will say nothing but your app might not work as you intended.

Just a pro tip :)

ozgurleo: Hi @keepitterron 
thanks a lot for your kind recommendations and reviews. 
today I am going to correct the points you mentioned and start to bonus sections. at least I can try:)

PR #238, opened 2018-09-14
week2 homework 

marwanaboudaoud:


week2/homework/src/index.js
```diff
@@ -1,3 +1,98 @@
 'use strict';
 
-// TODO: Write the homework code in this file
+let fs = require('fs');
+let command = process.argv[2];
+let index = process.argv[3];
```
J05J4: `index` is not the best variable name, since this is not always the index, in the case of `add` for example it is the actual text of the todo item...

week2/homework/src/index.js
```diff
@@ -1,3 +1,98 @@
...
+}
+
+function add(main) {
+    fs.appendFile('./to-do.txt', main + '\n', function (error) {
+        if (error)
+            console.log(error);
+    });
+}
+
+function Mydelete(index, reset) {
```
J05J4: Don't start your function with a capital letter: this should be `myDelete()`, or actually just `delete()`.

week2/homework/src/index.js
```diff
@@ -1,3 +1,98 @@
...
 
-// TODO: Write the homework code in this file
+let fs = require('fs');
+let command = process.argv[2];
+let index = process.argv[3];
+let item = process.argv[4];
+
+switch (command) {
+    case 'list':
+        listValue()
```
J05J4: Don't forget the semi colon `;`.

week2/homework/src/index.js
```diff
@@ -1,3 +1,98 @@
...
+function Mydelete(index, reset) {
+    fs.readFile('to-do.txt', 'utf8', function (error, data) {
+        if (error) {
+            console.log(error);
+        }
+        else {
+            let corrected = data.split('\n');
+            corrected.splice(index - 1, 1);
+            corrected = corrected.join('\n');
+            if (reset)
```
J05J4: Put curly braces behind an if-statement (it works how you did it now, but it is less readable).

week2/homework/src/index.js
```diff
@@ -1,3 +1,98 @@
...
+                corrected = "";
```
J05J4: It is good that you are making one function for similar functionality, you should however make this functionality more clear bij naming the function `removeItemOrClearList()` or something and don't pass it the number `1` as an argument, but just `true`, that will make it more clear what is going an. 
This is however a good mindset for a developer. You only have to improve it a little bit so that others also understand it when they read it.

PR #239, opened 2018-09-15
Sam week2

Samtabanja:
still working on it !


week2/homework/src/data.json
```diff
@@ -0,0 +1,6 @@
+{
+    "todo": {
+        "1": "shopping",
```
J05J4: When you are making an object, it makes more sense to use something like
```
{
    "id": 1,
    "text": "shopping"
}
```
That way you can use `todo.id` and `todo.text`. If you use the id as key, it is actually similar to an array.

week2/homework/src/index1.js
```diff
@@ -0,0 +1,91 @@
...
+    case 'remove':
+        remove(todo)
+        break;
+    case 'list':
+        todolist()
+        break;
+    case 'reset':
+        reset()
+        break;
+    case 'update':
```
J05J4: If I remove the third item with `node . remove 3` and then do `node . update 3 "asdasd"` a new item is created (with index 3), so the index is not updated if I remove an item. 
Maybe an easy solution could be to remove the id (number) altogether and just use an array, in that way the index updating is done automatically. In your json file you would have something like
```
todos: [
    { 
        "text": "shoping", 
        "done": false
    },
    {
        "text": "lunch",
        "done": false
    }
]
```
If you then want the first, you can do todos[0] and if you delete that one the second becomes the first automatically.

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
 'use strict';
 
 // TODO: Write the homework code in this file
+const fs = require('fs')
```
J05J4: add semi colon `;`

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
...
+    case 'reset':
+        reset()
+        break;
+    case 'update':
+        update(todo, todo2)
+        break;
+    case 'help':
+        help()
+        break;
+
```
J05J4: remove newlines

week2/homework/src/index1.js
```diff
@@ -0,0 +1,91 @@
+'use strict';
```
J05J4: index.js is set as `"main"` in package.json, so this file is not used then?

week2/homework/src/data.json
Samtabanja: thanks for this helpful feedbacks , all is done !

PR #240, opened 2018-09-15
Node js week2

WardMad:


week2/homework/src/index.js
```diff
@@ -1,3 +1,122 @@
...
+        arrayData.forEach((item) => {
+            split.push(item + '\n');
+
+            let replaceItem = split.toString().replace(arrayData[index - 1], newLine);
+
+            fs.writeFile('./data.txt', replaceItem.replace(/,/g, ''), function(error) {
+                if (error) {
+                    console.log(error);
+                }
+
```
J05J4: This was a difficult part, but you came up with a solution that is good! This is the kind of tasks you will be doing a lot as a programmer, so if it is frustrating you, please let me know which parts are difficult or cumbersome and maybe I can explain them further for you. Don't think a library (like express) is fixing these kind of problems for you: these are the basics and it is good to really onderstand them well!

PR #241, opened 2018-09-15
Barbaros Week2

whitepiolin:


week2/homework/src/index.js
```diff
@@ -1,3 +1,118 @@
 'use strict';
+let fs = require('fs');
```
keepitterron: Please consider using `const` over `let` as much as possible.
It helps defending you against bugs and prevents you changing stuff that shouldn't be changed.
You can [read more here](https://wesbos.com/let-vs-const/)

week2/homework/src/index.js
```diff
@@ -1,3 +1,118 @@
...
+let order = process.argv[2];
+let orderDefinition = process.argv[3];
+let orderDefinition2 = process.argv[4];
 
-// TODO: Write the homework code in this file
+fs.readFile('./toDo.json', 'utf8', function (error, data) {
```
keepitterron: Nesting everything inside a function is a bad practice.
It's difficult to understand what's going on and it will make testing these functions impossible.

week2/homework/src/index.js
```diff
@@ -1,3 +1,118 @@
...
+  if (error) {
+    console.log(error);
+  }
+  else {
```
keepitterron: This code can make use of [early return](https://dev.to/jenniferlynparsons/early-returns-in-javascript-5hfb) to make clear what you're doing.
Nesting everything inside an `else` makes the code harder to read. I use early returns all the time!

week2/homework/src/index.js
```diff
@@ -1,3 +1,118 @@
...
+        }
+        else {
+          console.log(data);
+        }
+      }
+      );
+    };
+
+    function list() {
+      function isEmpty(myToDoList) {
```
keepitterron: Again, do not nest functions inside functions, there's no reason to do it and the code becomes harder to follow and impossible to test.
the function `isEmpty` can live everywhere in this file and it still would work :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,118 @@
...
+      }
+    };
+
+    switch (order) {
+      case 'help':   // help shows help section
+        help();
+        break;
+      case undefined:   // no command shows help section
+        help();
+        break;
```
keepitterron: since the 2 `case`s above do the exact same thing you can rewrite this code as:
```js
case 'help':
case undefined:
  help();
  break;
```

week2/homework/src/index.js
```diff
@@ -1,3 +1,118 @@
 'use strict';
+let fs = require('fs');
+let order = process.argv[2];
+let orderDefinition = process.argv[3];
+let orderDefinition2 = process.argv[4];
```
keepitterron: Can you come up with a better name for these variables?
When reading it in the code I have no idea what order or order definition is. And even worse what orderdefinition2 is :)
[Naming stuff is very very hard](https://martinfowler.com/bliki/TwoHardThings.html), is the hardest stuff in development 😆 

> There are only two hard things in Computer Science: cache invalidation and naming things.

week2/homework/src/index.js
```diff
@@ -1,3 +1,118 @@
...
+        './toDo.json',
+        JSON.stringify(myToDoList),
+        (error) => {
+          if (error) { console.log(error); }
+        });
+    };
+
+    function remove(orderDefinition) {
+      if (orderDefinition > 0) {
+        delete myToDoList[orderDefinition];
```
keepitterron: In the help section it says `Removes a to-do item by its 1-base index.`, but this is not what's happening, am I wrong? It's using the 0-based index, so it doesn't do what the user expects.
Correct me if I'm wrong :)

week2/homework/src/index.js
keepitterron: Also, keep in mind that the `undefined` case is handled by the `default` case you have at the end of the `switch` :)

week2/homework/src/index.js
whitepiolin: yes i have read the topic and using const can be more logical. we do not need any update on those variables. Corrected

week2/homework/src/index.js
whitepiolin: Now, all functions are stated to alone. Corrected

week2/homework/src/index.js
whitepiolin: Corrected

week2/homework/src/index.js
whitepiolin: Understood, corrected

week2/homework/src/index.js
whitepiolin: I think that is not actually,  the code uses "const index = Object.keys(myToDoList).length + 1;" So it always starts from "1".  If we reset the list and add sth, we can see { '1': 'sth" }

keepitterron: You are absolutely right! my fault 🙇

week2/homework/src/index.js
whitepiolin: I have tried to change the program a bit but still some code are nested to other. Is it okey or should i need to change more?

keepitterron: can you git push your changes so i can see them? :)

whitepiolin:
@keepitterron  Hi Claudio, thank for your nice comments which are very helpful for me but at some points i have still need assistance. Could you please re examine the code according to my comments. Thanks again

week2/homework/src/index.js
whitepiolin: Corrected

week2/homework/src/help.txt
```diff
@@ -0,0 +1,10 @@
+
+===This program helps to create a to-do list.===
+
+Usable commands:
+
+list: 'Shows current to-dos',
+add: 'Adds a to-do item. All the words behind add are entered as 1 to-do item to the list.'
+remove: 'Removes a to-do item by its 1-base index.'
+reset: 'Removes all to-do items from the list.'
+update: 'Updates a to-do item with new text.'
```
keepitterron: You can write these command this way to make it more helpful for the end user:
`update [number] [todo]: Updates todo number [number] in the list with [todo]`
As a user when I check the help file I want to know how to use that command :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,134 @@
...
+
+function add(commandFirstItem) {
+  fs.readFile('./toDo.json', 'utf8', function (error, data) {
+    if (error) {
+      console.log(error);
+    }
+    else {
+      const myToDoList = JSON.parse(data);
+      const index = Object.keys(myToDoList).length + 1;
+      myToDoList[index] = commandFirstItem;
```
keepitterron: This works as intended, but you are using an object as an array.
There's nothing wrong with it in this case as it does what you want, but it's not the javascript way.

Try to rewrite the app so that the json looks like this:
```js
[
  {
    index: 1,
    description: 'Swim lesson 12.10.18 15:00 - 16:00'
  },
  {
    index: 2,
    description: 'Baseball lesson 13.10.18 15:00 - 16:00'
  },
];
```

So you have an array of objects to work with.

week2/homework/src/index.js
```diff
@@ -1,3 +1,134 @@
...
+          (error) => {
+            if (error) { console.log(error); }
+          });
+      }
+      else {
+        console.log('The parameter you entered as index is invalid');
+      }
+    }
+  });
+}
```
keepitterron: To clean up a bit more this function (and the other similar functions) you can use something I use all the time, [early returns](http://blog.timoxley.com/post/47041269194/avoid-else-return-early).
In a function when you use `return` the function will immediately stop and exit with what you return.
Example refactoring the callback function inside `fs.readFile`:
```js
// in case of error, stop immediately
if(error) {
  return console.log(error);
}
if (commandFirstItem < 1) {
  return console.log('The parameter you entered as index is invalid');
}

const myToDoList = JSON.parse(data);
delete myToDoList[parseInt(commandFirstItem)];
fs.writeFile(
  './toDo.json',
  JSON.stringify(myToDoList),
  (error) => {
    if (error) { console.log(error); }
  });
```

This is an advanced trick, I don't want to confuse you.
The code is exactly the same as what you wrote, but it's cleaner and easier to follow because the main logic is not indented :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,134 @@
...
+};
+
+function update(commandFirstItem, newData) {
+  fs.readFile('./toDo.json', 'utf8', function (error, data) {
+    if (error) {
+      console.log(error);
+    }
+    else {
+      const myToDoList = JSON.parse(data);
+      if (commandFirstItem > 0) {
```
keepitterron: Here you can apply the early return again.
The trick is: instead of thinking if `commandFirstItem` is what I want then do this, you need to think: if `commandFirstItem` is **not** what i want then stop and return an error :)

Does that makes sense?

week2/homework/src/help.txt
whitepiolin: Help list has been improved

week2/homework/src/index.js
whitepiolin: After editing, Now program works on arrays

week2/homework/src/index.js
whitepiolin: Yeah, I liked this view of early return. I have tried what I can do. I am curious about have I achieved it

whitepiolin: Yes I tried to edit it.

week2/homework/src/index.js
whitepiolin: Yes I tried to edit it.

whitepiolin:
@keepitterron Hi again Cladio, Really thanks about reviews. I tried to edit code. I will be waiting about your comment about my last changes. Since i want to learn this early return issue, so please make comments about is it done or not. Thank you so much

whitepiolin:
Yes, now i have tried also that style of return. It was a good learning session. Thanks for your efforts.

PR #242, opened 2018-09-15
Donal's week2

Donal86:


week2/homework/src/index.js
```diff
@@ -1,3 +1,125 @@
 'use strict';
 
 // TODO: Write the homework code in this file
+const fs = require('fs');
+const fileName = './index.txt';
+const defaultEncoding = 'UTF8';
+
+const program = require('commander');
```
keepitterron: 🎉

week2/homework/src/index.js
```diff
@@ -1,3 +1,125 @@
...
+    .description('remove a item from list')
+    .action(index => removeItem(index));
+program
+    .command('reset')
+    .description('removes all the items from the list')
+    .action(() => resetList());
+program
+    .command('update <index> <newItem>')
+    .description('update a item from the list')
+    .action((index, updatedTask) => updateItem(index, updatedTask));
```
keepitterron: This is and advanced topic and it's usually hard to understand at first. Don't worry if you don't get it or if it feels wrong. And feel free to ask questions if you have any :)

```js
action((index, updatedTask) => updateItem(index, updatedTask));
```
can be written as
```js
action(updateItem)
```

This is one of the beautiful things about javascript, you can pass functions as arguments to other functions.
In this case `action` will call the function for you, you can simplify this code by removing the intermediate function `.action(() => listItems())` ==> `.action(listItems)`

week2/homework/src/index.js
```diff
@@ -1,3 +1,125 @@
...
+            console.log(err);
+        }
+        else {
+            console.log(data);
+        }
+    });
+}
+
+function addItem(index) {
+    index = process.argv[3];
```
keepitterron: This function already receives the correct `index` from `commander`, you don't need to compute it yourself. You can remove this line and the app still works correctly :)
The same is valid for all other functions. `commander` parses `process.argv` for you so you don't have to deal with this error-prone stuff 🎉

week2/homework/src/index.js
```diff
@@ -1,3 +1,125 @@
...
+        if (err) {
+            console.log(err);
+        }
+        else {
+            console.log('Done!');
+        }
+    });
+}
+
+function updateItem(index, updatedTask) {
```
keepitterron: 🎉

week2/homework/src/index.js
```diff
@@ -1,3 +1,125 @@
...
+    });
+}
+
+function printHelp() {
+    console.log(`Usage: node index.js [options]
+HackYourFuture Node.js Week 2 - Homework To-Do App
+Options:
+  list               show all the tasks
+  add [to-do]        add task 
+  remove [to-do]     remove task
```
keepitterron: this instruction is unclear: you define it correctly as `command('remove <index>')` above. As a user I have no idea what `[to-do]` could be, it's much more clear if i read `remove [index]     remove task` so i know a number is expected :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,125 @@
...
+  update [to-do]     update a task from the list
```
keepitterron: the `update` command is also wrong. Don't I need to pass an index and a new todo item? :)

Donal86: Hi Claudio thank you for feedback it was very helpful to me. I have corrected my homework.
please advise.

Regards,

PR #244, opened 2018-09-16
Gaffari week2 homework

Gaffari02:


week2/homework/src/index.js
```diff
@@ -1,3 +1,81 @@
...
+    });
+}
+
+function removeItem(item) {
+    fs.readFile('./toDo.txt', 'utf8', function (error, data) {
+        if (error) {
+            console.error(error);
+        } else {
+            let itemRemove = data.split('\n');
+            itemRemove.splice(item, 1);
```
keepitterron: The assignment to remove an item says:
> Removes a to-do item by its 1-base index, e.g. to remove second item, execute:

```
node index.js remove 2
```

This function tells me I can remove an item by using `node . remove "todo item"` but it always remove the first item in the list.
Can you change it so that it works as requested? :)

PR #246, opened 2018-09-17
Node week3

mehmetscoskun:


mehmetscoskun:
After I finished the homework 1.0, I found a bug. That's why I made some changes.

week3/homework/src/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "index.js",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
```
J05J4: Your index is in `src/index.js`

week3/homework/src/manipulate.js
```diff
@@ -0,0 +1,91 @@
+'use strict';
+
+const fs = require('fs');
+const storage = './todos.json';
+
+function runTheFunction(func, param1, param2) {
```
J05J4: You can refactor this into `readData()` function, that only reads the data and returns it as JSON.
You can then call that function from for example an `add(item)` function (and all other functions):
```
function add(item) {
    const data = readData();
    const itemObj = {
        index: toDoList.length + 1,
        description: item
    };
    toDoList.push(itemObj);

    fs.writeFile(storage, JSON.stringify(toDoList), (error) => {
        if (error) { console.log(error); }
    });
}
```
In that way you can call the `add(item)` function directly from your route, so you would have in `src/index.js`:
```
app.post('todos, (request, response) => {
    manipulate.add(request.body.task);
    response
        .status(201)
        .json({ result: "your task is added!" });
});
```

week3/homework/src/manipulate.js
```diff
@@ -0,0 +1,91 @@
...
+
+    fs.writeFile(storage, JSON.stringify(toDoList), (error) => {
+        if (error) { console.log(error); }
+    })
+}
+
+function mark(data, index, type) {
+    const toDoList = JSON.parse(data);
+    toDoList[index - 1].done = type;
+    fs.writeFile(storage, JSON.stringify(toDoList), (error) => {
```
J05J4: This could also be a seperate function `writeData(toDoList)`, since you are using it in different places.

week3/homework/src/manipulate.js
```diff
@@ -0,0 +1,91 @@
...
+            } else {
+                console.log(err);
+            }
+        } else {
+            allFunctions(func, data, param1, param2);
+        }
+    });
+}
+
+function allFunctions(func, data, param1, param2) {
```
J05J4: I don't think you need this function.

mehmetscoskun:
Hi @J05J4 , thank you for your valuable comments. I guess I improved my code thanks to you. 
I added new functions. I did not know that I had to add first functions to the homework code. Now they are in.
I edited the `index.js` as `src/index.js` . But I dont actually understand why it is important. I guess I miss that part in the class.
I made my code more 'dry', for example now `fs.writeFile` function is not repetitive :)
At first, I could not handle async functions but now I understood better. I used `promise` and `await` . In this way, I guess it is more legible and effective.
Thank you for your all efforts. See you :)

PR #247, opened 2018-09-21
Week3 homework

sima-milli:
~I'm not satisfied with the code and I have a bug when adding new item to the json file if it is empty (exist but empty), I don't know how to fix it~
I added updated homework with all bugs fixed and more DRY code

### DIFF
--- a/week3/homework/src/index.js
+++ b/week3/homework/src/index.js
@@ -1,3 +1,153 @@
 'use strict';
 
-// TODO: Write the homework code in this file
+const fs = require('fs');
+const uuid = require('uuid/v4');
+const express = require('express');
+const bodyParser = require('body-parser');
+
+const app = express();
+app.use(bodyParser.json());
+
+// add new item to todo
+app.post('/add', (request, response) => {
+    if (!fs.existsSync('./todos.json')) {
+        fs.writeFile('todos.json', '', err => console.error(err));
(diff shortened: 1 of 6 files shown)


week3/homework/src/index.js
```diff
@@ -1,3 +1,56 @@
...
+const bodyParser = require('body-parser');
+const {
+    addItem,
+    removeItem,
+    listItems,
+    clearList,
+    updateItem,
+    markAsDone,
+    markAsNotDone
+} = require('./services');
```
J05J4: Nice clean imports!

week3/homework/src/index.js
```diff
@@ -1,3 +1,56 @@
...
+
+const app = express();
+app.use(bodyParser.json());
+
+// add new item to todo
+app.post('/add', (request, response) => {
```
J05J4: Since we are using the HTTP methods (`post`, `delete`, `get`, `put`) you don't really need the named endpoints: the post method is for creating a new item, so you could just `post` to `/todos` to create a new todo, or make a `get` request to `todos` to get all todos and make a `get` request to `todos/:id` to get a specific todo.

sima-milli: Thanks @J05J4  for your kind review 👍 :)

PR #249, opened 2018-09-21
Week3 Barbaros

whitepiolin:


week3/homework/src/index.js
```diff
@@ -1,3 +1,140 @@
...
-// TODO: Write the homework code in this file
+app.get('/list', function (request, response) {
+  fs.readFile('./toDo.json', 'utf8', function (error, data) {
+    if (error) {
+      console.log(error);
+      return;
+    }
+    const myToDoList = JSON.parse(data);
+    if (Object.values(myToDoList).length === 0) {
+      response.end('You have not any plan in your To Do List');
```
J05J4: Nicely handeling the edge cases!

week3/homework/src/index.js
```diff
@@ -1,3 +1,140 @@
...
+      return;
+    }
+    response.json(myToDoList);
+  });
+});
+
+app.get('/list/:id', function (request, response) {
+  fs.readFile('./toDo.json', 'utf8', function (error, data) {
```
J05J4: This functionality is used in multiple places, you should move the logic to a function, which can be called from the different places.

week3/homework/src/index.js
```diff
@@ -1,3 +1,140 @@
...
+    fs.writeFile('./toDo.json', JSON.stringify(myToDoList), function (error) {
+      if (error) {
+        console.error(error);
+      }
+    });
+    response.json(myToDoList);
+  });
+});
+
+app.delete('/delete', function (request, response) {
```
J05J4: Since you are already using the HTTP method `delete`, your path should just be '/todos' instead of '/delete'

week3/homework/src/index.js
```diff
@@ -1,3 +1,140 @@
...
+    if (error) {
+      console.log(error);
+      return;
+    }
+    const myToDoList = JSON.parse(data);
+    if (request.params.item < 0) {
+      response.end('The parameter you entered as index is invalid');
+      return;
+    }
+    if (request.params.item > Object.values(myToDoList).length) {
```
J05J4: Nicely handeling all the possible cases! 
I think that, since your myToDoList is an array of objects (`[{"index":0,"description":"SwimLesson"},{"index":1,"description":"DutchLesson"}]`) you could just use `myToDoList.length`...

week3/homework/src/index.js
```diff
@@ -1,3 +1,140 @@
...
+      response.end(`There is not any item in the list in the index of " ${request.params.item} "`)
+      return
+    }
+    myToDoList.splice(request.params.item, 1);
+    myToDoList.forEach(function (x) {
+      x.index = myToDoList.indexOf(x);
```
J05J4: Nice that you've thought about this! Very good! 
However, you are now resetting the index of an object in an array to the index of that array. You could remove the 'index' altogether from the objects and just use the array index. For example, if you have an array with objects in it like:
```
[
    {task: "do something"},
    {task: "do something else"},
    {task: "do something totally new"}
]
```
and want to get the first item of an array, you could use `myToDoList[0]` (you'll get `{task: "do something"}`).
If you remove this item from the array (`jsonData.splice(0,1)`), the second item becomes the first and the third becomes the second, but the id's of the array are already handeling that! 
If you now call `myToDoList[0]` you receive the new first (`{task: "do something else"}`)!

However, I think it is way more important that you thought about this, and found a solution to fix this!

week3/homework/src/index.js
```diff
@@ -1,3 +1,140 @@
...
+});
+
+app.post('/mark/:id/done', function (request, response) {
+  fs.readFile('./toDo.json', 'utf8', function (error, data) {
+    if (error) {
+      console.log(error);
+      return;
+    }
+    const myToDoList = JSON.parse(data);
+    myToDoList[request.params.id].done = "true";
```
J05J4: You could use the same kind of error handling here, if someone gives an id that does not exist as you did above...

week3/homework/src/index.js
```diff
@@ -1,3 +1,140 @@
...
+});
+
+app.delete('/mark/:id/done', function (request, response) {
+  fs.readFile('./toDo.json', 'utf8', function (error, data) {
+    if (error) {
+      console.log(error);
+      return;
+    }
+    const myToDoList = JSON.parse(data);
+    delete myToDoList[request.params.id].done;
```
J05J4: If you delete `done` the key and value are removed from the object. It is betere - since this value is of the type boolean - to set the value to false again.

PR #250, opened 2018-09-22
Genegr88 homework week3

genegr88:
finished homework

### DIFF
--- a/week3/homework/src/functions.js
+++ b/week3/homework/src/functions.js
@@ -1,0 +1,65 @@
+const fs = require('fs');
+
+function add(data, item) {
+    const to_do = JSON.parse(data);
+    const newTaskObject = {
+        index: to_do.length + 1,
+        description: item
+    };
+    to_do.push(newTaskObject);
+
+    fs.writeFile('./to-do.json', JSON.stringify(to_do), (error) => {
+        if (error) { console.log(error); }
+    });
+
+}
+
+function update(data, index, type) {
+    const to_do = JSON.parse(data);
+    to_do[index - 1].done = type;
+
+    fs.writeFile('./to-do.json', JSON.stringify(toDoList), (error) => {
+        if (error) { 
+            console.log(error); 
+        }
+    });
+
+}
+
+function showTask(data, index) {
+    const to_do = JSON.parse(data);
+
+    if (to_do.length === 0) {
+        return { "error": "Your list is empty" };
+    } else if (index > to_do.length) {
+        return { "error": "This task doesn't exist yet" };
+    } else {
+        return to_do[index - 1];
+    }
+}
+
+function reset() {
+    fs.writeFile('./to-do.json', '[]', (error) => {
+        if (error) { 
+            console.log(error); 
+        }
+    });;
+}
+function removeTask(data, index) {
+    const to_do = JSON.parse(data);
+    to_do.splice(index-1, 1)
+    fs.writeFile('./to-do.json', JSON.stringify(to_do)), (error) => {
+        if (error) {
+            console.log(error);
+        }
+    }
+}
+
+
+module.exports = {
+    add: add,
+    update: update,
+    showTask: showTask,
+    reset: reset,
+    removeTask: removeTask
+}
--- a/week3/homework/src/functions.js
+++ b/week3/homework/src/functions.js
@@ -2,15 +2,23 @@ const fs = require('fs');
 
 function add(data, item) {
     const to_do = JSON.parse(data);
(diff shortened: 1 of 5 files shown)


week3/homework/src/index.js
```diff
@@ -1,3 +1,99 @@
...
+const fs = require('fs');
+const express = require('express');
+const bodyParser = require('body-parser');
+const functions = require('./functions.js');
+const app = express();
+app.use(bodyParser.json());
+
+app.get('/todos/:id', (request, response) => {
+  const index = request.params.id;
+  fs.readFile('./src/to-do.json', 'utf8', (error, data) => {
```
J05J4: You are using `fs.readFile()` in multiple routes of your application: you should move this functionality to a function (like `getData()`) which you can call from the different routes.

week3/homework/src/functions.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+  });
+}
+
+function showTask(data, index) {
+  if (data.length === 0) {
+    return { 'error': 'Your list is empty' };
+  }
+  else if (index > data.length) {
+    return { 'error': "This task doesn't exist yet" };
```
J05J4: Nice error handling!

week3/homework/src/index.js
```diff
@@ -1,3 +1,99 @@
...
+        .status(500)
+        .json({ result: 'error' });
+      return;
+    }
+    response
+      .status(200)
+      .json(functions.showTask(todos, index));
+  });
+});
+app.delete('/todos', (request, response) => {
```
J05J4: It is more readable if you add a newline between different functions

week3/homework/src/functions.js
```diff
@@ -0,0 +1,69 @@
+const fs = require('fs');
+
+function add(data, item) {
+  const lastObject = data[data.length - 1];
+  let counter = 0;
+  if (lastObject === undefined) {
+    counter = 0;
+  }
+  else {
+    counter = parseInt(lastObject.id) + 1;
```
J05J4: If you remove your `index` from the JSON data, and just use te index of an array, you don't need a counter; you could just use `data[0]` to get the first and `data.length` to know at which position you need to add the new item. Now when you remove an item, the index isn't correct anymore. 
Try to fix this, and if you don't understand please let me know, because I will explain it to you: it is important to understand this since it is the basics hew to work with objects and arrays, you will use this a lot in a future job!

week3/homework/src/index.js
```diff
@@ -1,3 +1,99 @@
...
+    if (error) {
+      console.log(error);
+      response
+        .status(500)
+        .json({ result: 'error' });
+      return;
+    }
+    functions.update(todos, index, true);
+    response
+      .status(201)
```
J05J4: The task here is updated so the response code should be `200` (`201` is only for 'created')

week3/homework/src/functions.js
```diff
@@ -0,0 +1,69 @@
...
+  data.push(newTaskObject);
+
+  fs.writeFile('./to-do.json', JSON.stringify(data), (error) => {
+    if (error) throw error;
+    console.log(data);
+  });
+}
+
+function update(data, index, type) {
+  const todos = JSON.parse(data);
```
J05J4: The data was already parsed before you called this function

week3/homework/src/index.js
```diff
@@ -1,3 +1,99 @@
...
+    }
+    response
+      .status(200)
+      .json(functions.showTask(todos, index));
+  });
+});
+app.delete('/todos', (request, response) => {
+  functions.reset();
+  response
+    .status(301)
```
J05J4: `301` is the status code for moved permanently. Look at https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

week3/homework/src/index.js
```diff
@@ -1,3 +1,99 @@
...
+  fs.readFile('./src/to-do.json', 'utf8', (error, data) => {
+    const todos = data ? JSON.parse(data) : [];
+    if (error) {
+      console.log(error);
+      response
+        .status(500)
+        .json({ result: 'error' });
+      return;
+    }
+    functions.update(todos, index, false);
```
J05J4: Good thinking to reuse your update functionality!

week3/homework/src/functions.js
genegr88: Dont really understand

week3/homework/src/index.js
genegr88: Tried but then it doesnt work at all

PR #253, opened 2018-09-22
Week3

ozgurleo:
I could not achieve the 'markAsNotDone'. After I figure it out. I will recommit the homework shortly.


week3/homework/src/index.js
```diff
@@ -1,3 +1,75 @@
...
+});
+
+app.delete('/todos/:id', (req, res, next) => {
+  const { id } = req.params;
+
+  removeToDo(id)
+    .then(data => res.send(data))
+    .catch(err => next(err.message));
+});
+app.delete('/todos', (req, res, next) => {
```
J05J4: Make sure you give your functions some space by adding a line in between them, as you did above.

week3/homework/src/utilities.js
```diff
@@ -0,0 +1,108 @@
...
+const validateTask = (task) => {
+  const schema = {
+    id:Joi.number().required(),
+    description : Joi.string().required(),
+    done:Joi.boolean().required()
+  };
+  return Joi.validate(task, schema);
+};
+const appendToDo = async(item, res) => {
+  const data = await readToDoList();
```
J05J4: instead of `await` you can also use `fs.readFileSync` witch runs synchronously.

week3/homework/src/utilities.js
```diff
@@ -0,0 +1,108 @@
...
+      fs.appendFile(toDoFile, '[]');
+    };
+    fs.readFile(toDoFile, (err, data) => {
+      if (err) { return reject(err); };
+      return resolve(data);
+    });
+  });
+};
+function writeData(file, data) {
+  fs.watchFile(file, JSON.stringify(data), (error) => {
```
J05J4: Why are you using `watchFile()` instead of `writeFile()`?

week3/homework/src/utilities.js
```diff
@@ -0,0 +1,108 @@
...
+  };
+  toDoList.push(itemObj);
+  writeData(toDoFile, toDoList);
+  res.status(201);
+  res.json(itemObj);
+};
+const removeToDo = async(id, res) => {
+  const todoList = await readToDoList();
+  const removedTodo = todoList.find(task => task.id === id);
+  const index = todoList.indexOf(removedTodo);
```
J05J4: If you just use the index of an array (like in `todoList[`*0*`]`) you don't need the `id` in your object. 
You've now introduced a bug:
If I have a list of todo's:
```
[
    {
        index: 1,
        descritpion: "some description",
        done: false
    },
    {
        index: 2,
        descritpion: "some other description",
        done: false
    },
    {
        index: 3,
        descritpion: "some totally new description",
        done: false
    }
]
```
And I remove the second by making a `DELETE` request to `/todos/2`, it will look for `todoList[1]`, and remove that one. My list is now:
```
[
    {
        index: 1,
        descritpion: "some description",
        done: false
    },
    {
        index: 3,
        descritpion: "some totally new description",
        done: false
    }
]
```
When I now add a new item witch `appendToDo()`on line 34, it uses `toDoList.length + 1`, which is 3 now, os my list will now be:
```
[
    {
        index: 1,
        descritpion: "some description",
        done: false
    },
    {
        index: 3,
        descritpion: "some totally new description",
        done: false
    },
    {
        index: 3,
        descritpion: "something added after deleting an item",
        done: false
    }
]
```
You have two times `index: 3` now...
If you remove the index from the objects and just only use the `array` index, you will be fine!

week3/homework/src/utilities.js
```diff
@@ -0,0 +1,108 @@
...
+
+  if (typeof removedTodo === 'undefined') {
+    throw new Error('You have to enter a valid idea as number');
+  }
+
+  todoList.splice(index, 1);
+  await writeData(toDoFile, JSON.stringify(todoList, null, 2));
+  res.status(201);
```
J05J4: For removal we don't use status code `201`

week3/homework/src/utilities.js
```diff
@@ -0,0 +1,108 @@
...
+  res.json({'result': 'it is removed!'});
+};
+const clearToDoList = async(res) => {
+  await readToDoList();
+  writeData(toDoFile, JSON.stringify('[]', null, 2));
+  res.status(301);
```
J05J4: `301` is not the correct status code: 
https://stackoverflow.com/questions/2342579/http-status-code-for-update-and-delete

PR #255, opened 2018-09-23
Sam week3

Samtabanja:
homeWork week3


week3/homework/src/todo.js
```diff
@@ -0,0 +1,106 @@
...
+
+        return resolve(JSON.parse(data));
+      });
+    });
+  }
+
+  async readin(id) {
+    console.log(id)
+    const todos = await this.read();
+    const todo = todos.find(t => t.id == id);
```
J05J4: Preferably use `===` instead of `==`.

week3/homework/src/todo.js
```diff
@@ -0,0 +1,106 @@
...
+    return todo;
+
+  }
+
+  async deleteIn() {
+    const filteredTodos = []
+    return this._save(filteredTodos);
```
J05J4: You could just write 
```
async deleteIn() {
    return this._save([]);
}
```
Anyway `filteredTodos` is confusing as a name, since nothing is filtered here.

week3/homework/src/todo.js
```diff
@@ -0,0 +1,106 @@
...
+
+  async readin(id) {
+    console.log(id)
+    const todos = await this.read();
+    const todo = todos.find(t => t.id == id);
+    return todo;
+
+  }
+
+  async deleteIn() {
```
J05J4: What does `In` in the name `deleteIn()` mean?

Samtabanja: i really don't know ! ., while am busy with doing my homework i just used a random name , as i always do and i just forget to change it ... i change it for now hope it's better ...

week3/homework/src/index.js
```diff
@@ -1,3 +1,46 @@
...
+app.use(Express.json());
+
+app.post(`/${TODO_SLUG}`, createTodo.bind(null, todo));
+app.get(`/${TODO_SLUG}`, readTodos.bind(null, todo));
+app.put(`/${TODO_SLUG}/:id`, updateTodo.bind(null, todo));
+app.delete(`/${TODO_SLUG}/:id`, deleteTodo.bind(null, todo));
+
+
+app.get(`/${TODO_SLUG}/:id`, readSingalTodo.bind(null, todo));
+app.put(`/${TODO_SLUG}/:id/done`, markAsDone.bind(null, todo));
```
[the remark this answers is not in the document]
Samtabanja: well , i knew that i can use post in place of put ... but i thought both have the same effect in this situation ..

week3/homework/src/todo.js
Samtabanja: @J05J4 actually it takes 4 days to understand what's happening in the lecture , and 4 hours to solve the homework .... but i'm sure i understand what is happening in there and to be honest i'm still confused because it's something i can't build my self .
 thanks for your feedback ...

Samtabanja:
@J05J4 actually it takes 4 days to understand what's happening in the lecture , and 4 hours to solve the homework .... but i'm sure i understand what is happening in there and to be honest i'm still confused because it's something i can't build my self ...

week3/homework/src/index.js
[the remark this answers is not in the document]
J05J4: All methods have the same effect: the effect is created by the developer in the backend: you could use `DELETE` for the creation of a new todo item if you really wanted to. It is just a convention how to use them: `PUT` for creating a new item and `POST` for updating.

week3/homework/src/todo.js
J05J4: That's OK, it can happen! Just try to use meaningful names, so it is easier to understand the purpose.

week3/homework/src/actions/markAsDone.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+function markAsDone(todo, request, response) {
+  const id = request.params.id;
+  return todo.updateIn(id, true)
```
J05J4: Since `markAsDone()` and `markAsNotDone()` are exactly the same except for this `true` here and false in `markAsNotDone()`; this should be one function for example `toggleStatus(todo, status, {params}, response)`

PR #256, opened 2018-09-25
week3 homework

acimanx:
:tired_face:


week3/homework/src/controllers/crud.js
```diff
@@ -0,0 +1,129 @@
...
+    })
+    .catch(({ message }) => {
+      response.status(500);
+      response.json({ error: message });
+    });
+}
+function markAsDone(todo, { params }, response) {
+  const id = params.id;
+  return todo
+    .setTheFlag(id, true)
```
J05J4: Since `markAsDone()` and `markAsNotDone()` are exactly the same except for this `true` here and false in `markAsNotDone()`; this should be one function for example `toggleStatus(todo, status, {params}, response)`

PR #260, opened 2018-11-11
Finishes homework for week1

kawaraa:
Here is homework for Node.js week1


kawaraa:
Hi claudio @keepitterron, Thanks for your great comment!
I have implemented the homework week1 as you suggested.
let me know if I did correct!

PR #262, opened 2018-11-14
week1

salar01:


salar01:
Hello Claudio Semeraro. Thanks for your comments, it was useful and i learned a lot. I have changed my coeds as requested.

PR #263, opened 2018-11-14
NodeJS Week1 from Rabia

Damla78:


week1/homework/src/server.js
```diff
@@ -1,17 +1,40 @@
 'use strict';
 
+const url = require('url');
 const http = require('http');
+// const fs = require('fs');
```
keepitterron: when submitting work for code review, remember to always cleanup commented code :)

week1/homework/src/server.js
```diff
@@ -1,17 +1,40 @@
...
 
 /* `createServer` MUST return an instance of `http.Server` otherwise the tests
  * will fail.
  */
 function createServer(port) {
   let state = 10;
+  let newState = state;
```
keepitterron: this is amazing Rabia, and you're the only one to do this so far.
it's very important to not mutate the state, we will see why in the next lessons :)

week1/homework/src/server.js
```diff
@@ -1,17 +1,40 @@
...
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
-  });
+    const requestedUrl = url.parse(request.url, true);
```
keepitterron: this doesn't hurt and it works fine, but you don't need to parse the url for this particular case,
`request.url` would work just fine :)

week1/homework/src/server.js
```diff
@@ -1,17 +1,40 @@
...
+      response.writeHead(200, { 'Content-type': 'application/json' });
+      newState = newState + 1;
+      response.end(JSON.stringify({ state: newState }));
+    } else if (requestedUrl.pathname === '/subtract') {
+      response.writeHead(200, { 'Content-type': 'application/json' });
+      newState = newState - 1;
+      response.end(JSON.stringify({ state: newState }));
+    } else if (requestedUrl.pathname === '/reset') {
+      response.writeHead(200, { 'Content-type': 'application/json' });
+      newState = state;
```
keepitterron: 🎉

week1/homework/src/server.js
```diff
@@ -1,17 +1,40 @@
...
     // TODO: Write your homework code here
-  });
+    const requestedUrl = url.parse(request.url, true);
 
+    if (requestedUrl.pathname === '/state') {
+      response.writeHead(200, { 'Content-type': 'application/json' });
+      response.end(JSON.stringify({ state: newState }));
+    } else if (requestedUrl.pathname === '/add') {
+      response.writeHead(200, { 'Content-type': 'application/json' });
+      newState = newState + 1;
```
keepitterron: maybe you already know and just prefer it this way, you can make this shorter by doing `newState++` to increment a variable and `newState--` to decrement :)

week1/homework/src/server.js
Damla78: I cleaned.

week1/homework/src/server.js
Damla78: I changed.

week1/homework/src/server.js
```diff
@@ -1,20 +1,45 @@
...
+        break;
+      default:
+        writeRespond(404, { error: 'Not found' }, response);
+    }
   });
-
   return server;
 }
 
+function writeRespond(statuCode, stateMsg, response) {
```
keepitterron: great that you're passing the `response` to the function. This way this function is totally autonomous and it's easier to test (you'll learn about unit tests in one of the future lessons)

PR #272, opened 2018-11-23
my homework week2 node.js

RimSebai:


week2/homework/src/update.js
```diff
@@ -0,0 +1,14 @@
+const fs = require('fs');
+const data = require('./data');
+
+let d = [];
+const updateData = (a, b) =>
+  a == 1 || a == data.to_dos.length ?
```
keepitterron: this contains the same error as `removeData`, your condition checks if `a` is 1 OR if `a` is as long as the todo list length.
you want to check if `a` is greater than 0 AND `a` is not bigger than `data.to_dos.length`

```js
const data.to_dos = ['item1', 'item 2', 'item 3'];

// a == 1 || a == data.to_dos.length
updateData(1) // WORKS! 1 ==1 (true) OR 1 == 3 (false) (3 items in data.to_dos)
updateData(1) // DOES NOT WORK! 2 ==1 (false) OR 2 == 3 (false)
updateData(3) // WORKS! 3 ==1 (false) OR 3 == 3 (true)

// a > 0 && a <= data.to_dos.length
updateData(1) // WORKS! 1 > 0 (true) and 1 <= 3 (true) (3 items in data.to_dos)
updateData(1) // WORKS! 2 > 0 (true) and 2 <=3 (true)
updateData(3) // WORKS! 3 > 0 (true) and 3 <= 3 (true)
updateData(4) // DOES NOT WORK! 4 > 0 (true) and 4 <= 3 (false)
```

week2/homework/src/remove.js
```diff
@@ -0,0 +1,14 @@
+const fs = require('fs');
+const data = require('./data');
+
+let r = [];
+const removeData = a =>
+  a == 1 || a == data.to_dos.length ?
```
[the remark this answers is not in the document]
RimSebai: I use this function to remove a to-do item by its 1-base index and also to check if there is an item to remove or if the user enters unexpected input, and you are right about the condition I changed it.

week2/homework/src/update.js
RimSebai: thanks Claudio for your feedback!

PR #277, opened 2018-11-28
Rabia week3

Damla78:


ArcoMul:
Hi Rabia,

Good work! The application works as it should, and you abstracted all functionality in nice clear methods. 🙌 

A few things which would be good to correct though:
- You didn't add your packages to package.json. After I ran `npm install` and wanted to run your code it couldn't find `express` for example. You can solve this by running `npm install express` in the directory where your package.json is located (you might need to first delete the node_modules folder). The same counts for the `body-parser` module.
- When I add a todo (`POST /todo`) the `done` boolean is not added automatically. It would be good to set this to a default value (`done: false`)
- I also see that you don't use an `id` yet for your todos. You still work with the index of the array. In the lecture of week 3 is explained to use the `uuid` package to generate an id (https://github.com/HackYourFuture/Node.js/blob/master/week3/README.md#uuids). You will have to do this when adding a new todo to the array. Then the other methods need to be updated that they find a todo based on it's id instead of an index.
- And then the last thing to fix, is an easy one. Currently you export an array in `src/index.js` it's better to export an object so that you can use methods by name instead of find the right method in the array using an index. So instead of:
```
module.exports = [readToDoFile, addToDoFromHTML, updateToDoFromHTML, deleteToDoFromHTML, upDateDone, upDateNotDone];
// it's better to do:
module.exports = {
  readToDoFile,
  addToDoFromHTML,
  updateToDoFromHTML,
  deleteToDoFromHTML,
  upDateDone,
  upDateNotDone,
};
```
In you code you can then call the methods using:
`todoMethodList.readToDoFile(req.body.todo);`

I hope my comments are clear, otherwise make sure to write me here or on Slack.

Good luck with making the changes!

Arco

Damla78:
> * I also see that you don't use an `id` yet for your todos. You still work with the index of the array. In the lecture of week 3 is explained to use the `uuid` package to generate an id (https://github.com/HackYourFuture/Node.js/blob/master/week3/README.md#uuids). You will have to do this when adding a new todo to the array. Then the other methods need to be updated that they find a todo based on it's id instead of an index.

I did necessary changes. Please can you check my corrections?

week3/homework/src/index.js
```diff
@@ -1,3 +1,202 @@
...
+  newItem.id = uuidv4();
+  toDoList.push(newItem);
+  updateFile();
+  return 'New todo is added.';
+}
+//Index range starts from zero
+function updateToDoFromHTML(todoId, newToDo) {
+  const index = findIndexTodo(todoId);
+  if (index !== -1) {
+    toDoList[index].name = newToDo.name;
```
ArcoMul: Shouldn't this be `toDoList[index].description = newToDo.description;`?

week3/homework/src/server.js
```diff
@@ -0,0 +1,55 @@
...
+
+const app = express();
+app.use(bodyParser());
+
+const port = 3000;
+app.get('/todos', (req, res) => {
+  res.send(todoMethodList.readToDoFile());
+});
+app.get('/todos/:id', (req, res) => {
+  res.send(todoMethodList.readToDoFile(req.params.id));//todoMethodList[0](req.params.id));
```
ArcoMul: It currently doesn't return the todo, but by first converting the id to an index it should work:
`res.send(todoMethodList.readToDoFile(todoMethodList.findIndexTodo(req.params.id)));`
Right?

ArcoMul:
Hey Rabia,

Super good stuff! You did a lot of changes and (besides a few bugs) it works completely as it should! :-)

The only thing which is left is super small, but I wonder if you might have overlooked it, or that it maybe isn't clear? I'm talking about:

>You didn't add your packages to package.json. After I ran npm install and wanted to run your code it couldn't find express for example. You can solve this by running npm install express in the directory where your package.json is located (you might need to first delete the node_modules folder). The same counts for the body-parser module.

You can leave it as it is if you know how package.json works. But if you have any questions about it, let me know, because it is an important part of NodeJS.

Damla78:
> Hey Rabia,
> 
> Super good stuff! You did a lot of changes and (besides a few bugs) it works completely as it should! :-)
> 
> The only thing which is left is super small, but I wonder if you might have overlooked it, or that it maybe isn't clear? I'm talking about:
> 
> > You didn't add your packages to package.json. After I ran npm install and wanted to run your code it couldn't find express for example. You can solve this by running npm install express in the directory where your package.json is located (you might need to first delete the node_modules folder). The same counts for the body-parser module.
> 
> You can leave it as it is if you know how package.json works. But if you have any questions about it, let me know, because it is an important part of NodeJS.

I tried to do but I don't know what is the prb. If you are tomorrow in the course, we can do together.

PR #278, opened 2018-11-28
joudi week3 homework

joudi6:
working in progress on adding error handling part

### DIFF
--- a/week3/homework/package-lock.json
+++ b/week3/homework/package-lock.json
@@ -18,6 +18,11 @@
       "resolved": "https://registry.npmjs.org/array-flatten/-/array-flatten-1.1.1.tgz",
       "integrity": "sha1-ml9pkFGx5wczKPKgCJaLZOopVdI="
     },
+    "bignumber.js": {
+      "version": "4.1.0",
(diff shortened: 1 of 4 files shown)


ArcoMul:
Hi Joudi!

Great work, everything works as it should and your code is very clear!

Few things which you could improve, or take in mind next time:
- Currently you read the file once and then save the file every time you make a change to the list; this works perfectly and in this case there is nothing wrong with that solution. Overall it's better practice though to read the file again every time you need the data. In that way you are sure that the data you are working with is always exactly the same as the one you have in 'memory', saved in the `list` variable. When you only write the file (and never read it) there is a higher change that the data in the json file is not the same was what is in memory of your file. Or that the next time you run your program something crashes, because the file was not correctly written.
- `index.js` is now one long file with all the functionality. For readabilities sake and to keep a nice overview of your code it would be nice to split different functionality in modules. You could make a module called `todo.js` with methods like: `deleteTodo(list, id); createTodo(list, description); updateTodo(id, description); ...`

Good work! And let me know here or on Slack when something is not clear.

Arco

PR #279, opened 2018-11-28
Week3

RimSebai:


ArcoMul:
Hi @RimSebai 

Very clear and compact code! :-)

A few things to improve:
- The npm package you are using are not listed in your package.json. When someone (me in this case) wants to use your program they just want to run `npm install` to get all the right package. Now I had to figure out from your code which packages to install. You can easily fix this by running `npm install express` for example in the directory where your package.json is located (might be that you need to remove you node_modules directory first)
- Also you have an error / spelling mistake in your `POST /todos` route, which causes a crash
- And currently your `PUT /todos/:id` route only works when the id is also provided in the body of the post data. Although this is not necessarily wrong, it's not needed to repeat the id twice; once in the url and once in the post body

And some things to keep in mind for the next time:
- It's always good practice to split up your code in different modules. It makes the code more readable, mainly when trying to find where certain functionality is located. For example when I want to change something to the code where a todo is created it would make more sense to look in `todo.js` and then the `create(description)` method. Now I would have to search between the different routes.
- Currently you read the file once and then save the file every time you make a change to the todo list; this works perfectly and in this case there is nothing wrong with that solution. Overall it's better practice though to read the file again every time you need the data. In that way you are sure that the data you are working with is always exactly the same as the one you have in 'memory', saved in the `tod_os` variable. When you only write the file (and never read it) there is a higher change that the data in the json file is not the same was what is in memory of your file. Or that the next time you run your program something crashes, because the file was not correctly written.

Overall good work, and let me know if here or on Slack if you have any question at all!

Arco

PR #282, opened 2019-01-06
Reza- week1- An HTTP server that can add and subtract

ranjbarreza:


week1/homework/src/calculations.js
```diff
@@ -0,0 +1,26 @@
+function calculator(url, code, response) {
+  response.setHeader('Content-Type', 'text/html');
+  response.statusCode = code;
+  let state = 10;
+  if (url === '/') {
```
J05J4: You can use `if`, so this is a good solution, but just as a suggestion, you could also use `case switch`, which can make it more readable sometimes, in this example you would have: 
```
switch (url) {
    case '/':
        break;
    case '/state':
        break;
    case '/add':
        break;
    case '/subtract':
        break;
    case '/reset':
        state = 10;
        break;
    default:
        response.write('Not found.');
}
```

week1/homework/src/server.js
```diff
@@ -1,20 +1,33 @@
...
 /* `createServer` MUST return an instance of `http.Server` otherwise the tests
  * will fail.
  */
 function createServer(port) {
-  let state = 10;
+  // let state = 10;
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    if (request.url === '/') {
```
J05J4: You are going over the `path` twice now: once here and once in `calculations.js`. 
You should improve your code to go only over the `path` of the url once. Otherwise, if you ever want to add another route, you would have to do this in two places. That would be fine if you would have to do this in - say - a week, but if someone else would have to do this, or you have to do this in a year, or if this were part of a big application, it would be difficult, and one of the instances could be forgotten. 
This is part of the DRY (Don't Repeat Yourself) principle: you want to keep repeating code to a minimum to reduce the chance someone makes a mistake in one of the copies of the code. 
You can pass the `request.url` directly to `calculation.calculator()`.

ranjbarreza: Thanks Josja. I will go through my code again. Your comments are really useful.

ranjbarreza:
comments considered - DRY code - switch case

PR #283, opened 2019-01-08
node.js-cls18-wk1

mohamad-monir:


week1/homework/src/index.js
```diff
@@ -1,11 +1,9 @@
...
 
-const {
-  createServer
-} = require('./server');
+const { createServer } = require('./server');
 
 const PORT = 3000;
 
-createServer().listen(PORT, () => {
+createServer(PORT).listen(PORT, () => {
```
J05J4: Why are you passing `PORT` to the function `createServer()` as an argument?

week1/homework/src/server.js
```diff
@@ -10,10 +10,35 @@ function createServer(port) {
...
+        break;
+      default:
+        sendResponse(404, `not found`, response);
+    }
+  });
   return server;
 }
+function sendResponse(statusNum, content, response) {
+  response.statusCode = statusNum;
+  response.setHeader('content-type', 'application/json');
```
J05J4: The response you're sending is not completely of the `content-type` `application/json`, you should not mix content types: only use JSON or plain text!

week1/homework/src/server.js
```diff
@@ -10,11 +10,34 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    function sendResponse(statusNum, content) {
+      response.statusCode = statusNum;
+      response.writeHeader(200, { 'content-type': 'application/json' });
```
J05J4: You are now writing `statusNum` to `response.statusCode` and (regardless of the `statusNum` parameter) setting the header to `200` in `response.writeHeader()`.
You are mixing up two methods for setting headers: 
> `response.setHeader(name, value)`
> https://nodejs.org/api/http.html#http_response_setheader_name_value 

and

> `response.writeHead(statusCode[, statusMessage][, headers])`
> https://nodejs.org/api/http.html#http_response_writehead_statuscode_statusmessage_headers

Since you are already using `response.statusCode()`, `response.setHeader('content-type', 'application/json')` should be enough. 
```suggestion
      response.setHeader('content-type', 'application/json');
```

mohamad-monir: thank you for the feedback 
appreciated try to do what you asked me

PR #284, opened 2019-01-08
Yaacob wk1 homework

SignosRk:


ArcoMul:
Perfect execution Yaacob! Very readable and clean code 👏 

I could only find two possible improvements:
- Instead of using an if/if-else/else statement you could use a switch statement, that would make it even more readable
- The `getResponse` could be a function out of the scope of `createServer`, you would have to pass `response` as a parameter though. Currently the function `getRespone` is re-created everytime a requests comes in. This might cost unnecessary memory. So the benefit of moving the `getResponse` out of the `createServer` function is that it function only has to be created once in memory. _(Note here though that it's very likely that the NodeJS Javascript engine already does some optimization. So in the end, in this case, it might not really matter. Ask me on Sunday about this if you'd like to hear more)_

SignosRk:
@ArcoMul Thank you for the Feedback my brother, I shall work on the changes very soon.! Yes, I will change it to switch statements and move the getResponse out of the scope of createServer... and Yes... I want to understand better what have you just said about it... made me a bit curious to know more about... Thank you bro

SignosRk:
@ArcoMul the changes are done bro..  did a switch statement and also moved the `getResponse` out of the scoope of `createServer`...

PR #285, opened 2019-01-08
added week1 homework

Wael-Alhomsi:


ArcoMul:
Hi @Wael-Alhomsi, good job! Very clean and readable code! 👏

Two possible improvements:
- Instead of using an if/if-else/else statement you could use a switch statement, that would make it even more readable
- The `getResponse` could be a function out of the scope of `createServer`, you would have to pass `response` as a parameter though. Currently the function `getRespone` is re-created everytime a requests comes in. This might cost unnecessary memory. So the benefit of moving the `getResponse` out of the `createServer` function is that it function only has to be created once in memory. _(Note here though that it's very likely that the NodeJS Javascript engine already does some optimization. So in the end, in this case, it might not really matter. Ask me on Sunday about this if you'd like to hear more)_

Wael-Alhomsi:
Hey @ArcoMul !

Thank you so much for the review.
I have moved the function sendResponse out of the scope of the function createServer. In fact, at first I just wanted to stick to the skeleton original code provided, and wrote my code in the mentioned place.

All the best!

PR #286, opened 2019-01-09
Zerisenay solo home work Node.js week1-class 18 

ZerSolo:


ArcoMul:
Great work @ZerSolo, I see that you figured out how to use `JSON.stringify` :) Also nice that your sendResponse method is in it's own module! 💪 

Perfect execution, I see only one possible improvement:
```
sendResponse(response, {
  state: state
}, 200);
```
could also be written as:
```
sendResponse(response, {
  state
}, 200);
```
a little bit clearer to read!

See you Sunday!

PR #287, opened 2019-01-09
week1 homework

soumaiamamo:


week1/homework/src/server.js
```diff
@@ -2,14 +2,43 @@
...
+
+    function stateFunction(response, state) {
+
+      response.writeHead(200, { 'Content-Type': 'application/json' });
+      response.write(JSON.stringify({ state }));;
+      response.end();
+    }
+
+
+    if (request.url === "/state") {
```
J05J4: You can use `if`, so this is a good solution, but just as a suggestion, you could also use `case switch`, which can make it more readable sometimes, in this example you would have: 

```
switch (url) {
   case '/':
       break;
   case '/state':
       break;
   case '/add':
       break;
   case '/subtract':
       break;
   case '/reset':
       state = 10;
       break;
   default:
       response.write('Not found.');
}
```

week1/homework/src/server.js
```diff
@@ -2,14 +2,43 @@
...
+      stateFunction(response, state);
+    } else if (request.url === '/subtract') {
+      state -= 1;
+      stateFunction(response, state);
+    }
+    else if (request.url === "/reset") {
+
+      state = 10;
+      stateFunction(response, state);
+
```
J05J4: Try to have one style of coding you're now mixing the following:
```
if (condition) {
    code;

}
```
```
} else if (condition) {
    code;
}
```
```
} 
else if (condition) {
    code;
}
```
```
} else if (condition) {

    code;

}
```
They are all okay on their own, and it will work when you mix these styles, but it's more readable if you pick just one.
You should choose one style for newlines and formatting and stick with that.

PR #288, opened 2019-01-12
homework week 1

Kani103:


week1/homework/src/server.js
```diff
@@ -10,6 +10,36 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    const requestUrl = request.url;
+    if (request.url === '/state') {
```
J05J4: You can use `if`, so this is a good solution, but just as a suggestion, you could also use `case switch`, which can make it more readable sometimes, in this example you would have: 

```
switch (url) {
   case '/':
       break;
   case '/state':
       break;
   case '/add':
       break;
   case '/subtract':
       break;
   case '/reset':
       state = 10;
       break;
   default:
       response.write('Not found.');
}
```

week1/homework/src/server.js
```diff
@@ -10,6 +10,36 @@ function createServer(port) {
...
+      response.statusCode = 200;
+      response.setHeader('content-Type', 'application/json');
+      response.end(JSON.stringify({ state }));
+      // console.log(JSON.stringify({ state: state + 1 }));
```
J05J4: If you commit something, you should remove commented out code

week1/homework/src/server.js
```diff
@@ -10,6 +10,36 @@ function createServer(port) {
...
+    if (request.url === '/subtract') {
+      response.statusCode = 200;
+      response.setHeader('content-Type', 'application/json');
+      response.end(JSON.stringify({ state: --state }));
+    }
+    else
+    if (request.url === '/reset') {
+      response.statusCode = 200;
+      response.setHeader('content-Type', 'application/json');
+      response.end(JSON.stringify({ state: state = 10 }));
```
J05J4: You've written five times similar code: 
```
response.statusCode = 200;
response.setHeader('content-Type', 'application/json');
response.end(JSON.stringify({ state: state = 10 }));
```
The only thing changing in this code is the `statuscode` and the `state`. You should create a function if you are repeating the same code; the things that could be different in this block of code should be passed as arguments to that function. 
Refactor your code to make it more DRY (Don't Repeat Yourself)!

PR #290, opened 2019-01-15
Zerisenay Solomon Node.js HW Week2

ZerSolo:


week2/homework/src/index.js
```diff
@@ -1,3 +1,69 @@
...
+  let commandArg = parseInt(process.argv[3]);
+
+  if (!todosList) {
+    console.log('The TODO list file is empty');
+  } else if (commandArg > todosArray.length) {
+    console.log(`the requested line number ${commandArg} is empty`);
+  } else {
+    todosArray.forEach((element, index) => {
+      if (commandArg === index) {
+        let removedItem = todosArray.splice(index - 1, 1).join('\n');
```
J05J4:
```suggestion
        let removedItem = todosArray.splice(index - 1, 1);
```
You splice one item here, so joining one item by new line is not needed.

week2/homework/src/index.js
```diff
@@ -1,3 +1,69 @@
...
+        const restItems = todosArray.join('\n');
+        fs.writeFileSync('todos.txt', restItems);
+        console.log(removedItem);
```
J05J4: It would be more user friendly if the there was some more context here, like 
```suggestion
        console.log("The item '" + removedItem + "' is removed from the list");
```

week2/homework/src/index.js
```diff
@@ -1,3 +1,69 @@
...
+let reset = () => {
+  fs.writeFileSync('todos.txt', '');
+};
+
+
+let command = process.argv[2];
+
+switch (command) {
+  case ('help'):
+    console.log(fs.readFileSync('help.txt', 'utf8'));
```
J05J4: You've made a function you can call for every `case` in your `switch`, except for `help`, even though you are executing the same functionality twice. You should create a `help()` function as well, which you can call from the `'help'` `case` and the `default` `case`

ZerSolo:
Thank you Josya for your feedback and  encourgment. 
According to your feedback I have corrected my codes.More over, I have also done the bonas assignment.

PR #292, opened 2019-01-16
Class18 node wk2

mohamad-monir:


ArcoMul:
Hi @mohamad-monir,

First of all, very clean and readable code! I would be totally happy to work with this code.

For the sake of learning still some improvement points:
- **`const` instead of `let`**: You are currently using only `let` to declare variables and functions. As far as I can see you never put a new value in these variables, so they easily could be `const`'s. Making sure to use `const`'s when you can helps you to find mistakes earlier. You wouldn't want to assign a new value to a function variable for example. Currently -if you would do that- you wouldn't get an error though. While when the function is a `const` NodeJS would throw an error
- **Create pure functions:** A pure function is a function which returns the same value for the same input arguments. Your `add()` function isn't pure because it uses the global variable `command1`. So whenever calling the function it depends on this unknown global variable what the function would return. It would be better to pass the command as a function parameter into the function. In that way the function is more predictable. Read more about pure functions: https://en.wikipedia.org/wiki/Pure_function
- **Make functions DRY:** Currently you repeat yourself in some places. For example you are all the time reading the file using this line: `let file = fs.readFileSync('todo.txt', 'utf8');` Imagine that you would want to rename the todo.txt file to another name, todo-list.txt for example. You would have to replace the filename everywhere. It would be better to use a variable for the filename, or even better; to create a function which reads the file. `const file = readTodoFile();`. Read more about DRY: https://metova.com/dry-programming-practices/

I would be happy to review the changes you make.

See you Sunday!

ArcoMul:
We spoke about the pure function in class, right? Is it clear what it means?

In the add function of your code here it would mean to not use `command1` in the function but pass that value as a parameter:
```
const add = (targetedFile, value) => {
  readFiles(targetedFile);
  fs.appendFileSync(targetedFile, `${value}\n`);
  return value;
};

...

console.log(add('todo.txt', command1));
```

mohamad-monir:
Hi @ArcoMul I commit these changes before the class ... Thank you for clarifying what pure function is ...

PR #293, opened 2019-01-17
added homework week2

Wael-Alhomsi:


ArcoMul:
Hi @Wael-Alhomsi 

Great work, you code is very readable and well structured! 🙌 

Very good to see that you wrote your functions 'pure'. Meaning that the only data the function works with comes in through parameters. The same input always gives the same output. (https://en.wikipedia.org/wiki/Pure_function)

In the mindset of there is always something to learn, some points:
- **Use array for commands:** Currently you use `command` and `addition` as variable names for the commands. `addition` isn't a very descriptive name of course. I think it would be better to use an array for commands in that case, and pick the right index whenever needed. So `updateList(itemIndex, secondAddition);` would become: `updateList(itemIndex, commands[2]`. Then at least it is clear that what you are using is a command coming from the command line, in case of `addition` it could be any value.
- **Make functions DRY:** Currently you repeat yourself in some places. For example you are all the time reading the file using this line: `const list = fs.readFileSync('./data.txt', 'utf8');` Imagine that you would want to rename the data.txt file to another name, data-list.txt for example. You would have to replace the filename everywhere. It would be better to use a variable for the filename, or even better; to create a function which reads the file. `const list = readDataFile();`. Read more about DRY: https://metova.com/dry-programming-practices/

I would be happy to review the changes you make.

See you Sunday!

Wael-Alhomsi:
Hi @ArcoMul !
I really appreciate your kind review, I have updated my code accordingly. Thank you very much.
See you Sunday!

ArcoMul:
Looks good @Wael-Alhomsi! Might also be smart to put 'data.txt' in a `const` variable, then in case you would want to rename the file you only have to update that variable.

Wael-Alhomsi:
I have put  'data.txt' in a const variable. Thank you very much @ArcoMul !

PR #294, opened 2019-01-18
Zerisenay Solomon Week2 bonas assignment

ZerSolo:


ArcoMul:
Hi @ZerSolo 

Great that you picked up the bonus assignment!

It all looks very good. I see you divided up your code in different modules, used the command package in the right way. Also smart to put your help text in a seperate file.

Only saving JSON data to the file isn't correctly implemented. I think we already spoke about this in class. You are now stringify-ing and appending to a file. But you most likely want to `JSON.stringify()` a whole Javascript array or object and then write all of that to a file (not append).

Since reading and saving the JSON file is also included in the homework of week 3, feel free to leave this for what it is. Then I will look forward to see you implementation in the homework of week 3.

PR #295, opened 2019-01-19
added the bonus assignment homework

Wael-Alhomsi:


week2/homework/src/actions/removeFromList.js
```diff
@@ -0,0 +1,25 @@
...
+        toDoesArray.splice(index, 1);
+        writeF(JSON.stringify(toDoesArray));
+      } else {
+        writeF('');
+      }
+    } else {
+      console.log('The list has no corresponding "to-do element" to be removed.');
+    }
+  }
+}
```
ArcoMul: Hi @Wael-Alhomsi 

Great code! I really have nothing to complain about 😁 

Just a tip for readability's sake, the above function `removeFromList` has a lot of if-else statements. That can quickly get quite messy. Your current code is in that way a bit hard to figure out.

To solve this you can return earlier in the function. Apparently there is a name for this, I didn't know it myself either, but it's called a "Guard Clause". On this page they explain it quite well: https://medium.com/@scadge/if-statements-design-guard-clauses-might-be-all-you-need-67219a1a981a#33fa

When implementing this in your function it would look like this:
```js
function removeFromList(index) {
  const string = readF("./data.txt");

  // There is nothing to remove, console.log and return to stop function execution
  if (string === "") {
    console.log('The list has no corresponding "to-do element" to be removed.');
    return;
  }

  const toDoesArray = JSON.parse(string);

  // Again, nothing to remove, return
  if (index === 0 || index >= toDoesArray.length) {
    console.log('The list has no corresponding "to-do element" to be removed.');
    return;
  }

   // Just clear the file and stop execution
  if (toDoesArray.length === 1) {
    writeF("");
    return;
  }

  toDoesArray.splice(index, 1);
  writeF(JSON.stringify(toDoesArray));
}
```

More readable I think! :)

Wael-Alhomsi: Hi @ArcoMul !
Yes That's right. I have updated my code according to your tip and it is now much more readable.
Thanks and all the best :)

PR #296, opened 2019-01-19
Reza week2 - with bonus (JSON)

ranjbarreza:


ArcoMul:
Hi @ranjbarreza 

Great work! Readable code and great that you saved it in JSON already.

Two things which would be good to fix:

- **`const` instead of `let`**: You are currently using only `let` to declare variables and functions. As far as I can see you never put a new value in these variables, so they easily could be `const`'s. Making sure to use `const`'s when you can helps you to find mistakes earlier. You wouldn't want to assign a new value to a function variable for example. Currently -if you would do that- you wouldn't get an error though. While when the function is a `const` NodeJS would throw an error
- **Reset is broken:** The current implementation of the reset functionality causes the application to be broken. After I reset the todo list I can't add new items anymore. Try to figure out what's wrong (hint: it's an easy fix 😉)

PR #297, opened 2019-01-19
HW node week 2

soumaiamamo:


ArcoMul:
Hi @soumaiamamo 

Great work, the CLI application works as it should! 🙌

Regarding your code a few points:
- **Spacing & breaks:** Sometimes there are quite a lot of line-breaks in your file, between your functions and variables. Make sure too use not too many of them. Especially in you `add()` function it is hard to see where the function starts/stops and where the `removing()` function starts. Also in the `remove()` function you start the `fs.writeFile(...` part after a curly bracket of an if-statement. This is quite confusing 😉 Maybe also look into installing Prettier for VSCode, this extension takes care of formatting your code automatically: https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode 
- **Separate remove and reset:** The reset and remove functionality is now combined in one function. Which is good in the way that you don't repeat extra code. But it's also a bit confusing. If another programmer would work on this he/she would expect that the remove function just removes an item from the list. I think that in this case it would be better to create a separate `reset()` function. But alternatively you could rename the function to `removeOrReset()`.

PR #299, opened 2019-01-24
Zerisenay Solomon Node.js HW Week3-updated 

ZerSolo:


ZerSolo:
hi, Jos I have updated my codes and also cleaned the pull request.
 I changed the status code and some codes as well. 
But for the routes works all well with their error handling. Once an item added, deleted or updated it updates in the JSON file and gives also a response to the client. 
I am glad to hear from you if something is needed for correction.
with thanks,

ArcoMul:
Hi @ZerSolo,

Thanks for cleaning up the pull request by removing unnecessary files. You also removed the packages from package.json though, could you put those back? Those are needed when someone wants to use you application, to know what should be installed. Here you can see what changed: https://github.com/HackYourFuture/Node.js/compare/7e2c7bae1dd497be77f89b437f7493b312a69d0e..d957ee46fddf744266c35306d14a7e5f2ddbf625#diff-4e139b65c47201dc29a9fabb6ffec485

I also see that you made some changes on the code itself, great work! Did you do some good testing of you API though? As @J05J4  already said some of your routes don't function as they should. Of the few I tested I found these problems:
- `GET /todos` after `POST /todos` doesn't show the newly added todo (so in this case I first add a todo by doing a POST request, then I try to see all the todos by GET /todos, but then the one I just added doesn't show up in the list)
- Same for `GET /todos/:id` after `PUT /todos/:id`, it doesn't return the updated todo

Could you look at this and test your application another time? :-)

Arco

ZerSolo:
Hi Arco,
thanks for the feedback. I have tested all the requests in the postman and
they work perfectly.
I only have added the package.json  file.
I don't know what is missing?

Zerisenay

PR #301, opened 2019-01-24
Yaacob - express server from the class of nodeJS week3.

SignosRk:
This is the server we were setting the last sunday in the lecture.

### DIFF
--- a/week2/homework/src/index.js
+++ b/week2/homework/src/index.js
@@ -1,3 +1,40 @@
 'use strict';
 
 // TODO: Write the homework code in this file
+
+'use strict';
+// TODO: Write the homework code in this file
+let fs = require('fs');
+let command = process.argv[2];
+let value = process.argv[3];
+let index = parseInt(value);
+let updateTask = process.argv[4];
+
+const functionality = require('./functionality');
+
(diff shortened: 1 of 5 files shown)


week2/homework/src/index.js
```diff
@@ -1,3 +1,40 @@
...
+let updateTask = process.argv[4];
+
+const functionality = require('./functionality');
+
+switch (command) {
+  case 'list':
+    console.log(functionality.todoList());
+    break;
+  case 'add':
+    functionality.add(value);
```
J05J4: Because you have removed the `console.log`s from the functionality now, you are not giving any response to the user.

week2/homework/src/server.js
```diff
@@ -0,0 +1,37 @@
+const express = require('express');
+const bodyParser = require('body-parser');
+const port = 3000;
+const app = express();
+const functionality = require('./functionality');
+
+app.use(bodyParser.json());
+
+app.get('/list', (request, response) => {
+  console.log(request.body);
```
J05J4: You should remove these `console.log`s, since they are ending up on the server.

PR #302, opened 2019-01-24
added homework week3

Wael-Alhomsi:


week3/homework/src/actions.js
```diff
@@ -0,0 +1,105 @@
...
+    if (toDosArray[index].done === true) {
+      return 'The task is already done.';
+    } else {
+      toDosArray[index].done = true;
+      stringifyAndWrite(toDosArray);
+    }
+  }
+}
+
+function markUnDone(toDoId) {
```
J05J4: In general in programming if you have `markAsTrue()` and `markAsFlase()` you could replace these two functions with one "toggle" function: you are changing the state to the opposite state. In this specific case, you could do something like:
```
function toggleDone(toDoId) {
  const toDosArray = readAndParseData();
  const index = findIndex(toDoId, toDosArray);
  toDosArray[index].done = !toDosArray[index].done;
}
```

Wael-Alhomsi:
Hey @J05J4 !
I'm really happy to hear you positive feedback. Thank you very much :)

week3/homework/src/actions.js
Wael-Alhomsi: This is much easier and straightforward! 👍 @J05J4

PR #305, opened 2019-01-31
Reza week3- problems with DONE part

ranjbarreza:


ArcoMul:
Hi Reza,

Sadly your code doesn't run:

```
.\week3\homework\src\index.js:12
const tasks = JSON.parse(data);
                         ^
ReferenceError: data is not defined
```

Could you solve that?

Besides that I saw your comment `// not sure how to finish this part!`, could you post some question here on where you are stuck? Or which option you have considered? By having more info I can better estimate how I can help you ;-)

(Also make sure that the packages you use also end up in your package.json (by running `npm install express --save` for example)

So please let me know the questions you have, then I can help you with finishing the homework :-)

Arco

PR #308, opened 2019-02-18
 Utku / Class 19 / Homework Week 1

u61:


week1/homework/src/server.js
```diff
@@ -5,11 +5,40 @@ const http = require('http');
...
  * will fail.
  */
+
 function createServer(port) {
   let state = 10;
 
+  // Created a function in order to not repeating myself.
+  function showResponse(response) {
+    response.setHeader('Content-Type', 'application/json');
+    response.write(JSON.stringify({ state }));
```
keepitterron: as a tip: you are using a variable that exists somewhere else in the code.
this works fine, but you want your functions to be as independent as possible to reduce bugs and make them easier to understand.
I would pass `state` to the function 👍

u61:
Good Evening Claudio,

I have implemented my function to `default` section as well.
I did not require to include `response.statusCode = 404;` into the function and left it where it is.
I can change it if you want.

Appreciate for your reply and time.

PR #310, opened 2019-02-19
Homework of week1

ufukcalimli:


week1/homework/src/server.js
```diff
@@ -8,8 +8,36 @@ const http = require('http');
 function createServer(port) {
   let state = 10;
 
+  function respond(response) {
+    response.writeHead(200, {'Content-Type':'application/json'});
+    response.write(JSON.stringify({state}));
```
keepitterron: as a tip: you are using a variable that exists somewhere else in the code.
this works fine, but you want your functions to be as independent as possible to reduce bugs and make them easier to understand.
I would pass `state` to the function 👍

week1/homework/src/server.js
```diff
@@ -8,10 +8,10 @@ const http = require('http');
 function createServer(port) {
   let state = 10;
 
-  function respond(response) {
-    response.writeHead(200, {'Content-Type':'application/json'});
-    response.write(JSON.stringify({state}));
-    response.end();
+  function respond(status) {
```
keepitterron: why did you end up changing this parameter's name? It was fine :)
What I meant is to pass an additional parameter to the function to avoid reading a variable from outside the function's scope.

```js
function respond(response, stateObject) {
  response.writeHead(200, {'Content-Type':'application/json'});
  response.write(JSON.stringify(stateObject));
  response.end()
}
```

And you would call it like:
```js
respond(response, { status });
```

does it make sense?

ufukcalimli: Oh I thought the parameters name could make problem, that"s what I thought you meant.

PR #318, opened 2019-02-22
Elmira hw1 NodeJS

elmira202003:


week1/homework/src/server.js
```diff
@@ -7,9 +7,31 @@ const http = require('http');
  */
 function createServer(port) {
   let state = 10;
-
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    let FinalResponse = function(status, state) {
```
keepitterron: this is nit-picking, you are allowed doing what you did here and it works fine, but:
- for functions you always use `const` because you don't want to risk reassigning the variable
- you usually use this pattern with arrow functions, for old school functions you can simply do `function FinalResponse(status, state)`
- it's good practice to start variable names with a lower case letter `finalResponse` instead of `FinalResponse`. In JS is common to use uppercase letters for classes :)

PR #320, opened 2019-02-25
Ahmetbibi week2

ahmetbibi:


week2/homework/src/index.js
```diff
@@ -1,3 +1,62 @@
...
+    .command('add <newToDo>')
+    .description('to add a new to-do')
+    .action(newToDo => {
+      add(newToDo);
+    });
+
+  program
+    .command('remove <toDo>')
+    .description('to remove an existing to-do')
+    .action(toDo => {
```
saragerion: Not a big error, but I would suggest to call it "toDoNumber", same variable name used for the update function.
Naming consistency makes it more readable and neat.

week2/homework/src/functions.js
```diff
@@ -0,0 +1,104 @@
...
+      }
+    });
+  },
+
+  remove: index => {
+    fs.readFile('./to-do.txt', 'utf8', (err, data) => {
+      if (err) {
+        console.log(err);
+      } else {
+        if (index > data.split('\n').length) {
```
saragerion: Well done! But what happens if I try to run it with negative numbers, for example:
`node index.js remove -100`

ahmetbibi: Hello dear @saragerion I try to solve the minus problem and for this case, I think I solve. Thank you for your comments.

PR #323, opened 2019-03-01
HW2 - Node.js - fatih

kfth:


kfth:
> One minor but important tip. make sure you handle incorrect numerical inputs, like in the following scenarios:
> `node index.js remove -100`
> `node index.js update -100 update`

Thank you so much for your review. Incorrect numerical inputs like minus numbers in the arguments are handled by Comander library, if given an  unexpected input Commander throws "error: unknown option `-1'" error. I've also implemented that for 0 index since program bulit upon 1-base index

PR #325, opened 2019-03-01
Utku / Class 19 / Homework Week 2

u61:


week2/homework/src/index.js
```diff
@@ -1,3 +1,41 @@
...
+const help = require('./intro.js');
+const add = require('./add.js');
+const update = require('./update.js');
+const list = require('./list.js');
+const remove = require('./remove.js');
+const reset = require('./reset.js');
+
+/** ---------------------------------------
+     Commands
+ ---------------------------------------- */
```
saragerion: Great job with the commander here! Don't forget to handle the case when no command is specified (should display the help function)

week2/homework/src/reset.js
```diff
@@ -0,0 +1,30 @@
...
+ ---------------------------------------- */
+const reset = {
+  execute: () => {
+    const content = fs.readFileSync('To-Do Database.json', 'utf8');
+    switch (content) {
+      case '':
+        console.log('There is nothing to remove !');
+        break;
+      default:
+        fs.writeFile('To-Do Database.json', '', err => {
```
saragerion: In alternative of `''`, you could write an empty array: `'[]'`.
Doing that, you don't need to use fillEmptyDatabase in the add functionality

week2/homework/src/add.js
```diff
@@ -0,0 +1,40 @@
...
+     Base
+ ---------------------------------------- */
+const fs = require('fs');
+
+/** ---------------------------------------
+     Functions for Command
+ ---------------------------------------- */
+function fillEmptyDatabase() {
+  const content = fs.readFileSync('To-Do Database.json', 'utf8');
+  switch (content) {
```
saragerion: Switch is used when you have multiple scenario's (condition). If there is only 1 condition it makes more sense to use an `if` statement

week2/homework/src/update.js
```diff
@@ -0,0 +1,42 @@
...
+    switch (content) {
+      case '':
+        console.log('There is nothing to update !');
+        break;
+      default:
+        fs.readFile('To-Do Database.json', 'utf8', (err, readData) => {
+          if (err) throw err;
+          const parsedJSON = JSON.parse(readData);
+          const arrayLength = parsedJSON.length;
+          if (toDoNumber < 0 || toDoNumber > arrayLength - 1) {
```
saragerion: Watch out about this bit: `arrayLength - 1`.
What happens if I have only 1 item in the list, and I try to do:
node index.js update 1 "Random string"

week2/homework/src/index.js
u61: We could use 'node index.js' or 'node index.js help' for showing help section as per assignment. That's why I preffered index.js help. But, I've added commandless option as well.

week2/homework/src/reset.js
u61: But the client doesn't call 'reset' before each command. Also, the client always start with an empty database. Therefore, 'add' command always check for if it is an empty database.

week2/homework/src/add.js
u61: It's good to learn that. All switches have been switched to if statement.

week2/homework/src/update.js
u61: Yes. It's been corrected.

u61:
Hello @saragerion ,
I've tried to correct my code as per your comments.
Thank you for your review.

PR #326, opened 2019-03-02
Week2

AhmadAbdulhaleem:


week2/homework/src/index.js
```diff
@@ -1,3 +1,26 @@
...
+  case 'list':
+    method.list();
+    break;
+  case 'remove':
+    method.remove(param);
+    break;
+  case 'reset':
+    method.reset();
+    break;
+  default:
```
saragerion: When you run `node index.js` or `node index.js` the output looks like the following:
```
{ add: [Function: add],
  list: [Function: list],
  remove: [Function: remove],
  reset: [Function: reset] }
```
Can you change it so that it will display some instructions of usage for each commands?

week2/homework/src/arguments.js
```diff
@@ -0,0 +1,9 @@
+let arguments = process.argv;
+const file = arguments.slice(2);
+
+const retrieve = argument => arguments.includes(argument);
```
saragerion: What's the usage of this retrieve function?

week2/homework/src/methods.js
```diff
@@ -0,0 +1,48 @@
...
+  },
+
+  list: () => {
+    fs.readFile(path, { encoding: 'utf-8' }, (err, data) => {
+      console.log(data);
+    });
+  },
+
+  remove: row => {
+    const removeLines = (data, lines = []) => {
```
saragerion: What happens if we try to remove an item with index 1000, or -1000? Consider the edge cases

AhmadHaleeem: thank you 
I solved all issues

PR #327, opened 2019-03-02
Soolaf Week2 homework

soolaf:


week2/homework/src/index.js
```diff
@@ -1,3 +1,95 @@
...
+
+switch (command) {
+  case 'list':
+    viewTasks();
+    break;
+  case 'add':
+    addTask();
+    break;
+  case 'help':
+  case undefined:
```
keepitterron: `case undefined` is the same as `default`. you should move `case help` all the way down before `default` as the assignment says you should return the `help` section if you call the app without any command :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,95 @@
 'use strict';
 
-// TODO: Write the homework code in this file
+const fs = require('fs');
+let command = process.argv[2];
+let value = process.argv[3];
+let listFile = fs.readFileSync('./data.json', 'utf8');
+let listArr = JSON.parse(listFile);
```
keepitterron: one thing you should try - it helps a lot with thinking about the application - is use better names.
`listArr` doesn't help understanding what that is, `todoList` is maybe a better name.
same with `listFile`, maybe `todoJson` will be easier to understand what this is when you look at this code again in 3 months :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,95 @@
...
+    });
+  }
+}
+
+function addTask() {
+  if (process.argv.length < 4) {
+    console.log('check the help menu by typing help');
+  } else {
+    let taskLine = '';
+    for (let i = 3; i <= process.argv.length - 1; i++) {
```
keepitterron: you don't need to do this, but it's a clever trick!
when adding a todo, you should call it with `node index.js add "my todo"`. this way you don't have to check for all the `argv` after the command, you can just get that one `argv`.

week2/homework/src/index.js
```diff
@@ -1,3 +1,95 @@
...
+    listArr.push(taskLine);
+    fs.writeFile('./data.json', JSON.stringify(listArr, null, 2), err => {
+      if (err) throw err;
+    });
+  }
+}
+
+function removeTask(i) {
+  if (i > listArr.length || i == 0) {
+    console.log('not existed task type list to view existed tasks (starting index is 1) ');
```
keepitterron: i like that you specify the starting index 👍

week2/homework/src/index.js
```diff
@@ -1,3 +1,95 @@
...
+  case 'update':
+    updateTasks(value);
+    break;
+
+  default:
+    console.log('wrong command check the help menu by typing help');
+}
+
+function viewTasks() {
+  if (!listArr || listArr.length == 0) {
```
keepitterron: good practice to always check 🎉

week2/homework/src/index.js
```diff
@@ -1,3 +1,95 @@
...
+  fs.writeFile('./data.json', listArr, err => {
+    if (err) throw err;
+  });
+}
+
+function updateTasks(i) {
+  if (i > listArr.length || i == 0) {
+    console.log('not existed task type list to view existed tasks (starting index is 1)');
+  } else if (process.argv.length < 5) {
+    console.log(' check the help menu by typing help');
```
keepitterron: it would be good to display the help section directly here, no? so that the user doesn't have to retype the command with help :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,95 @@
...
+  }
+  let taskLine = '';
+  for (let i = 4; i <= process.argv.length - 1; i++) {
+    taskLine += process.argv[i] + ' ';
+  }
+  listArr.splice(i - 1, 1, taskLine);
+  fs.writeFile('./data.json', JSON.stringify(listArr, null, 2), err => {
```
keepitterron: you should turn this into a function, so you don't have to repeat the same code again and again. what do you think?

```js
function saveJsonData(data) {
  fs.writeFile('./data.json', JSON.stringify(data, null, 2), err => {
    if (err) throw err;
  })
}
```

PR #328, opened 2019-03-03
Week2

YaseirAlkhwalda:


week2/homework/src/index.js
```diff
@@ -1,3 +1,19 @@
 'use strict';
 
-// TODO: Write the homework code in this file
+const actions = require('./actions');
+
+const args = process.argv;
+
+args.splice(0, 2);
+
+const retrieve = argument => args.includes(argument);
```
keepitterron: what happens if i don't pass any command? `node index.js`?
I should get the `help` menu. but i get nothing instead :)

week2/homework/src/actions.js
```diff
@@ -0,0 +1,78 @@
...
+        });
+      }
+    });
+  },
+  help: () => {
+    console.log(`
+  CLI todo List App 
+  use: node index [options]
+  Options:
+    add...............To add an item: (ex: node index add Hi)
```
keepitterron: i really like that you provided an example on how to use the commands! 💃

week2/homework/src/actions.js
```diff
@@ -0,0 +1,78 @@
...
+const actions = {
+  add: element => {
+    readFile('todo.txt', 'utf-8').then(data => {
+      if (data === '') {
+        fs.appendFile('todo.txt', element, err => {
+          if (err) throw err;
+          console.log('an item has been added ...');
+        });
+      } else {
+        fs.appendFile('todo.txt', `\n${element}`, err => {
```
keepitterron: you're repeating the same function twice, that's a sign that something can be extracted.
you can either write a function that does all that and you pass it `element` or `element/n`
```js
const saveTodoItem = element => {
  fs.appendFile('todo.txt', element, err => {
    if (err) throw err;
    console.log('an item has been added ...');
  });
}
```

and you call it like 
```js
if(data === '') saveTodoItem(element)
else saveTodoItem(`\n${element}`)
```

OR you can simply create an intermediate variable to append
```js
const line = data === '' ? element : `\n${element}`;
```

week2/homework/src/actions.js
```diff
@@ -0,0 +1,78 @@
...
+  remove: number => {
+    readFile('todo.txt', 'utf-8').then(data => {
+      if (data === '') {
+        console.log(`There is no items to remove...`);
+      } else {
+        const elements = data.split('\n');
+        elements.map((element, index) => {
+          if (number == index + 1) {
+            elements.splice(number - 1, 1);
+            fs.writeFile('todo.txt', elements.join('\n'), err => {
```
keepitterron: why did you promisify readFile but not writeFile? just curious :)

YaseirAlkhwalda:
> good job @YaseirAlkhwalda !! I've left a couple of comments :)

week2/homework/src/actions.js
YaseirAlkhwalda: thanksssssss a lot ^_^

week2/homework/src/index.js
YaseirAlkhwalda: Hi @keepitterron  Thanks for your feedback, I fixed it.
if (retrieve('add')) actions.add(process.argv[1]);
else if (retrieve('remove')) actions.remove(process.argv[1]);
else if (retrieve('list')) actions.list(process.argv);
else if (retrieve('reset')) actions.reset(process.argv);
else if (retrieve('help')) actions.help(process.argv);
else actions.help(process.argv);

week2/homework/src/actions.js
YaseirAlkhwalda: :clap:

PR #345, opened 2019-05-12
RamazanInce NodeJs HW1

RamazanInce:


week1/homework/src/server.js
```diff
@@ -5,13 +5,38 @@ const http = require('http');
...
  */
+
 function createServer(port) {
   let state = 10;
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
-  });
+    if (request.url.startsWith('/state')) {
+      response.writeHead(200, { 'Content-Type': 'application/json' });
```
yash-kapila: Hi Ramazan - I am going to be a bit strict starting now as we are getting into advanced topics and it is important that we get the basics right.

Over here - if you look closely, the `{ 'Content-Type': 'application/json' }` is being repeated in all the if-elseif-else scenarios. This means all the response types are the same. Thus, you can move this object out in a constant and then use that variable in the `writeHead`.

week1/homework/src/server.js
```diff
@@ -5,13 +5,38 @@ const http = require('http');
...
+      response.write(JSON.stringify({ state: state }));
+    }
+ else if (request.url === '/add') {
```
yash-kapila: Hi Ramazan - the indentation of the `else-if` blocks aren't correct. Could you please check?

week1/homework/src/server.js
```diff
@@ -5,13 +5,38 @@ const http = require('http');
...
+      state = state + 1;
+      response.writeHead(200, { 'Content-Type': 'application/json' });
+      response.write(JSON.stringify({ state: state }));
+    }
+ else if (request.url === '/subtract') {
+      state = state - 1;
+      response.writeHead(200, { 'Content-Type': 'application/json' });
+      response.write(JSON.stringify({ state: state }));
```
yash-kapila: Over here - you can make use of the ES6 shorthand syntax instead of writing as { state: state }. Please have a look here - https://alligator.io/js/object-property-shorthand-es6/

week1/homework/src/server.js
```diff
@@ -5,13 +5,38 @@ const http = require('http');
...
+    }
+ else if (request.url === '/reset') {
+      state = 10;
+      response.writeHead(200, { 'Content-Type': 'application/json' });
+      response.write(JSON.stringify({ state: state }));
+    }
+ else {
+      response.writeHead(404, { 'Content-Type': 'application/json' });
+      // obj = { error: 'Not found' };
```
yash-kapila: We can remove commented out code here.

week1/homework/src/server.js
```diff
@@ -5,13 +5,38 @@ const http = require('http');
...
  * will fail.
  */
+
 function createServer(port) {
   let state = 10;
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
-  });
+    if (request.url.startsWith('/state')) {
```
yash-kapila: Any specific reason why you went with `startsWith` method here but used triple equals in other cases?

week1/homework/src/server.js
RamazanInce: I am doing in a correct way but Visual Studio auto correct it like that.

week1/homework/src/server.js
RamazanInce: Thanks to strict @yash-kapila  you can be more strict on it.
I have done it.

week1/homework/src/server.js
RamazanInce: Oh, I am not aware of that, thanks for it. It is ok now.

week1/homework/src/server.js
RamazanInce: I did it. thanks

week1/homework/src/server.js
RamazanInce: Oh sorry, I was trying to do something different then I forgt to change it back.

PR #346, opened 2019-05-13
node js week1

alaashhadat:


week1/homework/src/server.js
```diff
@@ -9,7 +8,35 @@ function createServer(port) {
...
+      case '/subtract':
+        state -= 1;
+        response.write(JSON.stringify({ state }));
+        break;
+      case '/reset':
+        state = 10;
+        response.write(JSON.stringify({ state }));
+        break;
+      default:
+        if (request.url === '/') {
```
eidosam: > Any other URL results in an error with status code 404 and response body in JSON format

https://github.com/HackYourFuture/Node.js/blob/master/week1/homework/README.md#endpoints-to-implement

So I think root also should yield to 404

But in case you want to accept calls to '/' then you can combine the two cases ('/' and '/state') together

    switch (request.url) {
      case '/':
      case '/state':
        response.write(JSON.stringify({ state }));
        break;
      // ... the rest

Ref. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#Multi-case_-_single_operation

week1/homework/src/server.js
```diff
@@ -9,7 +8,35 @@ function createServer(port) {
...
+          response.write(JSON.stringify({ state }));
+        }
+        // eslint-disable-next-line indent
```
eidosam: If there is linting error, better to fix it instead of disabling the rule

PR #348, opened 2019-05-13
Wadeea node homework week1

WadeeaKiwan:


WadeeaKiwan:
> Very well done Wadeea! Nice to use [JSDoc](https://jsdoc.app/index.html) for documenting your functions, but make sure to use the right types, like `number` instead of `integer` and `http.ServerResponse` or `object` instead of `response`.
> 
> Also please don't commit any changes not related to your solution.

Many thanks @eidosam for your time and feedback. The changes are done.

I'll consider the commits in the next homeworks.

PR #349, opened 2019-05-14
Enes Node.Js Week 1 HW

eneserdek:
…e Is  Required in server.Js

### DIFF
--- a/week1/homework/src/urlExtension.js
+++ b/week1/homework/src/urlExtension.js
@@ -1,0 +1,29 @@
+module.exports = {
+  myStateFunction: function(request, response, state) {
+    if (request.url === `/state`) {
+      response.writeHead(200, { 'Content-Type': 'application/json' });
+      response.write(JSON.stringify({ state }));
+    }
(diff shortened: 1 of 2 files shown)


eneserdek:
I am trying to make my code better. So, please make your comments after Thursday 23:59.

week1/homework/src/urlExtension.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable */
```
eidosam: Please don't disable Eslint, instead try to fix the lint errors/warnings if exist.

week1/homework/src/urlExtension.js
```diff
@@ -0,0 +1,30 @@
...
+module.exports = {
+  myStateFunction: function(request, response, state) {
+    function writeHeadAndStringify(code, value) {
+      response.writeHead(code, { 'Content-Type': 'application/json' });
+      typeof value === 'number'
```
eidosam: Wouldn't be easier if you test the `code` instead of the type of `value`?

PR #354, opened 2019-05-17
I made my homework

Kani103:


week1/homework/src/server.js
```diff
@@ -10,10 +10,35 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    const requestUrl = request.url;
+    switch (requestUrl) {
+      case '/state':
+      response.statusCode = 200;
```
eidosam: This will be applied only on this case, so the following cases (except for the error case - default) will take the default status-code, which is luckily 200.

week1/homework/src/server.js
```diff
@@ -10,10 +10,35 @@ function createServer(port) {
...
+      response.setHeader('content-Type', 'application/json');
```
eidosam: This line can be moved to before the switch block, because it is repeated in all cases.

PR #357, opened 2019-05-19
semih-week1

sigma21:


sigma21:
Thanks a lot @eidosam for the feedback. I'll implement your advice about keeping my code dry. Thanks again!

PR #359, opened 2019-05-20
Ramazan Ince NodeJs HW2

RamazanInce:


week2/homework/src/index.js
```diff
@@ -1,3 +1,68 @@
 'use strict';
 
-// TODO: Write the homework code in this file
+let fs = require('fs');
+let args = process.argv.slice(2);
+// console.log(args);
```
eidosam: Commented code can be removed

week2/homework/src/index.js
```diff
@@ -1,3 +1,68 @@
...
+
+let command = args[0];
+let todoItem = args[1];
+
```
eidosam: `fs`, `args`, `command` and `todoItem` are not re-assigned, so they have to be declared with `const`

week2/homework/src/index.js
```diff
@@ -1,3 +1,68 @@
...
+if (command === undefined || 'help') {
```
eidosam: This is always evaluated to `true`. You can read it this way

    if ((command === undefined) || 'help') {
      // ...
    }

It should be re-written

    if (command === undefined || command === 'help') {
      // ...
    }

Or ` (!command || command === 'help') `

week2/homework/src/index.js
```diff
@@ -1,3 +1,68 @@
...
+      } else console.error(error);
+    } else {
+      console.log(todoList);
+    }
+  });
+}
+
+function add(todoItem) {
+  fs.appendFile('./todoList.txt', todoItem + '\n', error => {
+    console.error(error);
```
eidosam: The `error` should be checked, and logged only if it is present. So the reporting to user will be proper.
Like

    if(error) return console.error(error);
    return console.log('Todo item is added');

week2/homework/src/index.js
```diff
@@ -1,3 +1,68 @@
...
+  });
+}
+
+function remove(index) {
+  fs.readFile('./todoList.txt', 'utf8', (err, data) => {
+    if (err) throw err;
+    let newList = data.split('\n');
```
eidosam: `newList` is not re-assigned, so it can be declared with `const`

week2/homework/src/index.js
```diff
@@ -1,3 +1,68 @@
...
+    newList.splice(index - 1, 1);
+    fs.writeFile('./todoList.txt', newList.join('\n'), err => {
+      if (err) throw err;
+    });
+  });
+}
+
+function reset(todoItem) {
+  fs.writeFile('./todoList.txt', '', function(err) {
```
eidosam: Use arrow-functions where possible.

week2/homework/src/index.js
```diff
@@ -1,3 +1,68 @@
...
+    if (err) throw err;
+    console.log('Reset!');
+  });
+}
+
+function help(todoItem) {
+  fs.readFile('./help.txt', 'utf8', function(err, data) {
```
eidosam: Use arrow-functions where possible.

week2/homework/src/index.js
RamazanInce: I did not try that one, I did it in the last step, Thanks for the correction.

week2/homework/src/index.js
RamazanInce: that is ok, thanks

wouterkleijn:
Thanks @eidosam for addressing the comments. Thanks @RamazanInce for implementing the necessary changes :)

PR #362, opened 2019-05-23
Node HM week2

Muhannad1989:


week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
...
+      console.log('invalid number');
+    }
+  } catch (err) {
+    console.error(err);
+  }
+}
+
+function write(arr) {
+  return new Promise((resolve, reject) =>
+    fs.writeFile(fileName, `${arr.join('\n')}`, (err, data) => (err ? reject(err) : resolve(data))),
```
eidosam: No need for template-strings here, it can be written simply `arr.join('\n')`

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
...
+  return new Promise((resolve, reject) =>
+    fs.appendFile(fileName, `${text.join(' ')}\n`, (err, data) =>
+      err ? reject(err) : resolve(data),
+    ),
+  );
+}
+
+function list() {
+  return new Promise(resolve =>
+    fs.readFile(fileName, 'utf8', (err, data) => resolve(err ? '' : data)),
```
eidosam: Maybe you need to `reject` the error like above?

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
...
+To-Do App
+
+Options:
+
+  read            read all to-dos
+  write [to-do]   add to-do
+  reset           remove all to-dos
+  remove          remove a specific item in to-dos by passing a index number
+  help            show this help text
+  `);
```
eidosam: Commands listed in help don't match with the actual commands. `read` => `list`, `write` => `add`

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
...
+  help();
+} else if (command === 'list') {
+  list().then(data => console.log(`To-Dos:\n${data}`));
+} else if (command === 'remove') {
+  remove(number);
+} else if (command === 'reset') {
+  reset().then(() => console.log('All to-do items has been deleted'));
+} else if (command === 'update') {
+  update(number, newItem);
+}
```
eidosam: You need also to handle the case when user enter an invalid command.

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
...
+
+async function update(index, text) {
+  try {
+    const data = await list();
+    let arr = data.split('\n');
```
eidosam: `arr` is nor re-assigned, so it can be declared with `const`

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
...
+    write(arr);
```
eidosam: Not needed. But in case you want to keep it, it should be `await`'ed, because `write` returns a promise.

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
...
+    if (index <= arr.length && index > 0) {
+      arr.splice(index - 1, 1, text);
+      write(arr).then(console.log('an item has been removed'));
```
eidosam: an item has been ~removed~ modified

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
...
+    }
+  } catch (err) {
+    console.error(err);
+  }
+}
+
+async function remove(index) {
+  try {
+    const data = await list();
+    let arr = data.split('\n');
```
eidosam: `const`

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
...
+    if (index <= arr.length && index > 0) {
+      arr.splice(index - 1, 1);
+      write(arr).then(console.log('an item has been modified'));
```
eidosam: an item has been ~modified~ removed

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
...
-// TODO: Write the homework code in this file
+const fs = require('fs');
+const fileName = `${__dirname}/todoList.txt`;
+
+const command = process.argv[2];
+const args = process.argv.slice(3);
+
+const newItem = process.argv.slice(4);
+
+const number = parseInt(args);
```
eidosam: `args` is an array, so you should select an item of this array to parse it as int, but because JavaScript is merciful, it still works, the first element of the array will be converted to int. Try not to rely on JavaScript flexibility and write more comprehensive code.

week2/homework/src/index.js
```diff
@@ -1,3 +1,97 @@
 'use strict';
 
-// TODO: Write the homework code in this file
+const fs = require('fs');
+const fileName = `${__dirname}/todoList.txt`;
+
+const command = process.argv[2];
+const args = process.argv.slice(3);
+
+const newItem = process.argv.slice(4);
```
eidosam: Variables `args` and `newItem` are not used in all commands, so better to declare them when needed only.

week2/homework/src/index.js
Muhannad1989: Hi Samir, Very big thanks for you. Each time you correct my homework i learn new something in coding.
I've modified all issues you mentioned here, you can have a look if you like. 
Thanks again.

Muhannad1989:
Thank you Samir for your feedback and for checking the code again.

wouterkleijn:
Thanks @eidosam for the comments. Great job @Muhannad1989

PR #363, opened 2019-05-23
command line interface app

ghifo:
The remove function does not work as supposed yet, I am still working on it.

### DIFF
--- a/week2/homework/src/index.js
+++ b/week2/homework/src/index.js
@@ -5,7 +5,6 @@ The user must be able to add, remove and list to-dos.
 The user must be able to remove all to-dos at once.
 The functions are: 
   add();
-  help();
   list();
   remove();
   update();
@@ -13,19 +12,19 @@ The functions are:
   help();
   It should control what happens if user enters unexpected input, e.g. remove -100
 */
-let fs = require('fs');
-let args = process.argv.slice(2);
+const fs = require('fs');
+const error = require('./error');
+
+const args = process.argv.slice(2);
 // shows args starting from the . in the command line
-let command = args[0];
+const command = args[0];
 // the 1st arg is the command that calls the meant function
-let todoItem = args[1];
+const todoItem = args[1];
 // the 2nd arg is the item we are dealing with of a number of its position
-let newVal = args[2];
+const newVal = args[2];
 // the 3rd arg is the item's updated value
 if (command === 'add') {
   add(todoItem);
-} else if (command === 'help') {
-  help();
 } else if (command === 'list') {
   list();
 } else if (command === 'remove') {
@@ -34,18 +33,14 @@ if (command === 'add') {
   update(todoItem, newVal);
 } else if (command === 'reset') {
   reset();
-} else if (command === 'help' || !command) {
+} else if (command === 'help' || !command || todoItem <= 0 || todoItem > todoList.length) {
   help();
 }
 
 function list() {
   fs.readFile('./src/todoList.txt', 'utf8', (error, todoList) => {
     if (error) {
-      if (error.code === 'ENOENT') {
-        console.log('no data found');
-      } else {
-        console.log(error);
-      }
+      error.e404(error);
     } else {
       console.log(todoList);
     }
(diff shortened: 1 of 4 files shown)


week2/homework/src/index.js
```diff
@@ -1,3 +1,161 @@
...
+      } else {
+        console.log(error);
+      }
+    } else {
+      if (num < 0 || num > todoList.length) {
+        console.log('Please write a valid number');
+      } else {
+        const splitList = todoList.split('\n');
+        // make the list an array of lines
+        const newList = splitList.filter((elem, index) => index !== num - 1).join('\n');
```
eidosam: May [splice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) serves this case better than `filter`?

week2/homework/src/index.js
```diff
@@ -1,3 +1,161 @@
...
+        console.log(newList);
+        fs.writeFile('./src/todoList.txt', newList, error => console.error(error));
```
eidosam: Print the error message only when there is. Same way you did in `add` function, lines (57-61

week2/homework/src/index.js
```diff
@@ -1,3 +1,161 @@
...
+      } else {
+        console.log(error);
+      }
+    } else {
+      if (num < 0 || num > todoList.length) {
+        console.log('Please write a valid number');
+      } else {
+        const splitList = todoList.split('\n');
+        console.log(num);
+        console.log(newVal);
```
eidosam: Do you need to show the user these messages?

week2/homework/src/index.js
```diff
@@ -1,3 +1,161 @@
...
+
+        // make the list an array of lines
+        const newList = splitList
+          .map((elem, index) => (index === num - 1 ? newVal : elem))
+          .join('\n');
+        // console.log(newList);
+        fs.writeFile('./src/todoList.txt', newList, error => console.error(error));
```
eidosam: Same note about handling/printing the errors.

week2/homework/src/index.js
```diff
@@ -1,3 +1,161 @@
...
+      }
+    }
+  });
+}
+function reset() {
+  fs.writeFile('./src/todoList.txt', '', (error, todoList) => {
+    if (error) {
+      if (error.code === 'ENOENT') {
+        console.log('no data found');
```
eidosam: Error `ENOENT` occurs in `fs.writeFile` when the parent directory of the file is not present. So this message is invalid.

week2/homework/src/index.js
```diff
@@ -1,3 +1,161 @@
...
+        console.log(error);
+      }
+    } else {
+      console.log(`Rest!
+      Nothing to be done in the list`);
+    }
+  });
+}
+function help() {
+  fs.readFile('./src/todoList.txt', 'utf8', error => {
```
eidosam: Help command is only for printing the usage guide, nothing else. So no need to check the existence of the data file.

week2/homework/src/index.js
ghifo: Shall I write the error this way, or no need at all for using error hier?
 if (error) {
      console.log(error);
    }

week2/homework/src/index.js
ghifo: Is there any problem of using filter here?
I was advised to use it :)

week2/homework/src/index.js
eidosam: That should be good enough. As you suggested

    if (error) {
        console.log(error);
    } else {
        console.log(`Rest!
            Nothing to be done in the list`);
    }

week2/homework/src/index.js
eidosam: Not a problem, but it is different use case. Usually `filter` is used to filter data on the content of the array, not the index of the elements.

week2/homework/src/index.js
ghifo: Thanks

PR #369, opened 2019-05-26
semih week-2

sigma21:


week2/homework/src/index.js
```diff
@@ -1,3 +1,66 @@
...
 
 // TODO: Write the homework code in this file
+
+let fs = require('fs');
+
+let args = process.argv.slice(2);
+
+let command = args[0];
+let todoItem = args[1];
+
```
eidosam: All the above variables are not re-assigned, so `const` should be used to declare them.

week2/homework/src/index.js
```diff
@@ -1,3 +1,66 @@
...
+
+
+if(command === 'add') {
+  add(todoItem);
+} else if (command === 'help'){
+  help();
```
eidosam: Try to run `node . help`; you will get an error

    ReferenceError: helpText is not defined

So you need to move the declaration of `helpText` to before the function call.

week2/homework/src/index.js
```diff
@@ -1,3 +1,66 @@
...
+} else if (command === 'list') {
+  list(todoItem);
+} else if (command === 'remove') {
+  remove(todoItem);
+};
+
```
eidosam: No command or invalid one should be also covered.

week2/homework/src/index.js
```diff
@@ -1,3 +1,66 @@
...
+      console.log(error);
+    } else {
+      console.log(todoList);
+    };
+  });
+};
+
+function add(todoItem){
+  fs.appendFile('./todoList.txt', todoItem + '\n', (error) => {
+    console.error(error);
```
eidosam: You should also check the error and print it only if there is. Also better to report to user the status, like  

    if(error) {
        console.error(error);
    } else {
        console.log('Item is added.');
    }

Like what you did in `remove` function.

week2/homework/src/index.js
```diff
@@ -1,3 +1,66 @@
...
+  });
+};
+
+function remove(todoItem){
+  fs.readFile('./todoList.txt', 'utf8', (err, data) => {
+    if (err) {
+      throw err;
+    } else {      
+      let dataArray = data.split('\n');
```
eidosam: `dataArray` is not re-assigned, use `const`

wouterkleijn:
Thanks @eidosam for the comments. Please @sigma21  can you make sure to implement the feedback?

wouterkleijn:
@sigma21 Any updates?

sigma21:
@wouterkleijn @eidosam thanks for the feedback!
I've just updated the PR according to your comments.

wouterkleijn:
Super, guys - thanks a lot :)

PR #372, opened 2019-05-30
Akif_NodeW3_Homework

Akif2709:


week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+    const parsedList = JSON.parse(todoList);
+    if (error) {
+      if (error.code === 'EN0ENT') {
+        console.log('No data found');
+      } else {
+        console.error(error);
+        response.status(500).send('some problem.');
+      }
+    } else if (request.params.id < parsedList.length) {
+      const todoItem = parsedList[++request.params.id]['text'];
```
eidosam: You don't want to change the ID, so don't use `++` or `--`. Also you want to change from 1-based to 0-based order, so can use `request.params.id-1` instead of `++request.params.id`
Also you may need to check zero or negative ID, and you need to accept the last element

        } else if (0 < request.params.id <= parsedList.length) {

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+      response.json(todoItem);
+    } else {
+      response.json('Please enter a valid id');
```
eidosam: When the user provides invalid input, the response code should be 4xx, in you case `400` which means Bad Request, also the message you are sending back is not JSON, so you use `response.send`;  So `response.status(400).send('Please enter a valid id')`

Even better if you split it into two cases;
- When ID <= 0 it means Bad Request (400)
- When ID > Todo-List-Size  it means Not Found (404)

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+  });
+}
+
+function add(request, response) {
+  fs.readFile('./todoList.json', 'utf8', (error, data) => {
+    if (error) {
+      console.log(error);
```
eidosam: You need always to respond to client; otherwise the request will keep hanging until it timeout

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+      let todos = JSON.parse(data);
```
eidosam: `const`

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+      const todo = { text: request.params.text, done: false };
+      todos.push(todo);
+      let dataString = JSON.stringify(todos);
```
eidosam: `const`

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+      fs.writeFileSync('./todoList.json', dataString, 'utf8');
```
eidosam: You need to handle the error it might happens on file-write, and send the proper status code accordingly.

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+  });
+}
+
+function clearTodos(request, response) {
+  fs.writeFile('./todoList.json', '[]', err => {
+    if (err) {
+      console.log(err);
+    }
+    response.send('All todos is deleted!');
```
eidosam: This status message will be sent even if there is an error, I don't think this is what you want.

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+}
+
+function markAsDone(request, response) {
+  fs.readFile('./todoList.json', 'utf8', (error, data) => {
+    if (error) {
+      console.log(error);
```
eidosam: You need to respond here also.

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+      let todos = JSON.parse(data);
```
eidosam: `const`

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+      todos[--request.params.id]['done'] = true;
```
eidosam: Similar to the comment in `readTodo`; use `request.params.id-1`, also you need to check the existence of the element, like what you did earlier.

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+      let dataString = JSON.stringify(todos);
```
eidosam: `const`

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+      fs.writeFileSync('./todoList.json', dataString, 'utf8');
```
eidosam: Similar to the note in `add` function; you need to handle the errors.

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+      console.log(error);
+    } else {
+      let todos = JSON.parse(data);
+      todos[--request.params.id]['done'] = false;
+      let dataString = JSON.stringify(todos);
+      fs.writeFileSync('./todoList.json', dataString, 'utf8');
+      response.send(todos);
+    }
+  });
+}
```
eidosam: Same notes in `markAsDone` applies here

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+
+function remove(request, response) {
+  fs.readFile('./todoList.json', 'utf8', (error, data) => {
+    if (error) {
+      console.log(error);
```
eidosam: Do not forget to respond to client

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+      let todos = JSON.parse(data);
+      todos.splice(--request.params.id, 1);
+      let dataString = JSON.stringify(todos);
+      fs.writeFileSync('./todoList.json', dataString, 'utf8');
+      response.send(todos);
+    }
```
eidosam: Same notes in `markAsDone` applies here

week3/homework/src/actions.js
```diff
@@ -0,0 +1,100 @@
...
+      console.log(error);
+    } else {
+      let todos = JSON.parse(data);
+      todos[--request.params.id]['text'] = request.params.newText;
+      let dataString = JSON.stringify(todos);
+      fs.writeFileSync('./todoList.json', dataString, 'utf8');
+      response.send(todos);
+    }
+  });
+}
```
eidosam: And here, same notes in `markAsDone`

week3/homework/src/index.js
```diff
@@ -1,3 +1,25 @@
...
+} = require('./actions');
 
-// TODO: Write the homework code in this file
+const app = express();
+app.use(bodyParser.json());
+
+app.get('/todos/:id', readTodo);
+app.delete('/todos', clearTodos);
+app.post('/todos/add/:text', add);
+app.post('/todos/:id/done', markAsDone);
```
eidosam: Shouldn't be `put` method here?

week3/homework/src/index.js
```diff
@@ -1,3 +1,25 @@
...
+app.delete('/todos/:id/done', markAsNotDone);
```
eidosam: Also here, shouldn't be `put` method?

Akif2709: Thanks for feedback Samir 👍

Akif2709: I made the corrections and push it but it doesn't seem here yet. No idea

wouterkleijn:
Thanks @eidosam for the feedback and @Akif2709 for addressing the suggestions made by Samir :)

week3/homework/src/actions.js
```diff
@@ -0,0 +1,122 @@
...
+
+function clearTodos(request, response) {
+  fs.writeFile('./todoList.json', '[]', err => {
+    if (err) {
+      console.log(err);
+      response.status(500).send('some problem.');
```
eidosam: You need `return` here; `return response.status(500).send('some problem.');` so it doesn't try to send response twice in case of errors.

PR #377, opened 2019-05-31
Enes Node.js HMW 3

eneserdek:


week3/homework/src/methodList.js
```diff
@@ -0,0 +1,150 @@
...
+        resolve(data);
+      }
+    });
+  });
+}
+
+function writeFile(path, data) {
+  return new Promise((resolve, reject) => {
+    fs.writeFile(path, JSON.stringify(data), 'utf8', error => {
+      reject(error);
```
eidosam: This will always reject, instead you have to check the error and reject or resolve accordingly; like

    if(error) return reject(error);
    return resolve();

week3/homework/src/methodList.js
```diff
@@ -0,0 +1,150 @@
...
+
+async function markAsDone(number, path, response) {
+  try {
+    await readFile(path)
+      .then(data => {
+        const todoList = JSON.parse(data);
+        const todo = todoList.find(todo => {
+          return todo.id == number;
+        });
+        todo.done = true;
```
eidosam: Like what you did in `readToDo` function, you need to check if the requested todo exists.

week3/homework/src/methodList.js
```diff
@@ -0,0 +1,150 @@
...
+}
+async function markAsNotDone(number, path, response) {
+  try {
+    await readFile(path)
+      .then(data => {
+        const todoList = JSON.parse(data);
+        const todo = todoList.find(todo => {
+          return todo.id == number;
+        });
+        todo.done = false;
```
eidosam: Same remark; you need to check if the requested todo exists.

week3/homework/src/methodList.js
```diff
@@ -0,0 +1,150 @@
...
+        writeFile(path, todoList);
+      })
+      .catch(error => {
+        console.log(error);
```
eidosam: Don't forget to send response indicating the error

week3/homework/src/methodList.js
```diff
@@ -0,0 +1,150 @@
...
+      });
+    listToDo(path, response);
+  } catch {
+    response.status(404).send({ error: 'there is an error' });
+  }
+}
+
+function clearToDos(path, response) {
+  writeFile(path, []).then(response.json('You now have a brand new todo list'))
```
eidosam: When you send plain text, use `response.send` instead of `response.json`

week3/homework/src/methodList.js
```diff
@@ -0,0 +1,150 @@
...
+  .catch(error=>response.status(404).send({ error: 'there is an error' });)
+}
+
+async function removeToDo(todoItem, path, response) {
+  try {
+    await readFile(path).then(data => {
+      const todoList = JSON.parse(data);
+      const filteredOutElement = todoItem - 1;
```
eidosam: Since you are using the ID of the todo in the previous functions to find it in the list, not the index/order of the item; you should also here be consistent and use the same method here.

week3/homework/src/methodList.js
```diff
@@ -0,0 +1,150 @@
...
+  } catch {
+    response.status(404).send({ error: 'there is an error' });
+  }
+}
+
+async function updateToDo(todoItem, todo, path, response) {
+  try {
+    await readFile(path).then(data => {
+      const todoList = JSON.parse(data);
+      todoList[--todoItem].name = todo.item;
```
eidosam: Same here, you should us the ID of the todo, not the index

        const currentTodo = todoList.find(todo => {
          return todo.id === todoItem;
        });
        currentTodo.name = todo.item

wouterkleijn:
Thanks @eidosam for your feedback. Please @eneserdek  Can you have a look at the comments and address them?

week3/homework/src/methodList.js
eneserdek: I removed id references since i used uuid.

eidosam:
Thanks Enes!

PR #379, opened 2019-06-01
Mohammad-Darwesh-homework-node-week3

mohammadnaefaldarwesh:


mohammadnaefaldarwesh:
> Good job Mohammad! The `actions` are very well done, but the routes needs to be re-written in ExpressJS way, where you don't check the path and method yourself, let ExpressJS do it for you.
> 
> For example
> 
> ```
> const { add, get, getOne } = require('./actions');
> router.get('/', (req, res) => res.status(200).json(get()));
> router.get('/:id', (req, res) => res.status(200).json(getOne(req.params.id)));
> router.post('/', (req, res) => res.status(201).json(add(req.body.task)));
> ```
> 
> This is not complete, just translation of your current implementation to Express routes, you still need to handle the errors and client's bad requests.

Hello @eidosam ,
thank you for your feed back .
actually the express router is already checking  the path and the method in index.js and the main function only diced  which status and response will it be send and in the same function i handle the error and i tested it and it is work perfectly.......if you see that i still need to make it as your example please let me know.
Best regards,

eidosam:
> the main function only for deciding which status and response will it be send and handling the error

This is exactly routing, which is the what Express Router does, and how you did use it in `index.js`

> in this way i do not need to make eight different functions and i do not need to handle the same error eight times

Let's try this

    const main = {
        getTodos: (req, res) => res.status(200).json(get()),
        getTodo: (req, res) => res.status(200).json(getOne(req.params.id)),
        addTodo: (req, res) => res.status(201).json(add(req.body.task)),
        markTodoDone: (req, res) => res.status(201).json(done(req.params.id)),
        markTodoNotDone: (req, res) => res.status(201).json(notDone(req.params.id)),
        removeAllTodos: (req, res) => res.status(201).json(removeAll()),
        removeTodo: (req, res) => res.status(201).json(remove(req.params.id) && get()),
        updateTodo: (req, res) => res.status(201).json(update(req.params.id, req.body.task)),
        errorHandler: (error, req, res) => res.status(400) && res.json(error)
    };

(which is also just translation of your current implementation)
I think this way make it a little bit more cleaner, and you don't have to create variables will not be used in all calls.

> and i tested it and it is work perfectly

I'm pretty sure you did, but it is not only about working code, we are aiming at working and clean code.


Re: Error handling, please have a look at https://expressjs.com/en/guide/error-handling.html

mohammadnaefaldarwesh:
> > the main function only for deciding which status and response will it be send and handling the error
> 
> This is exactly routing, which is the what Express Router does, and how you did use it in `index.js`
> 
> > in this way i do not need to make eight different functions and i do not need to handle the same error eight times
> 
> Let's try this
> 
> ```
> const main = {
>     getTodos: (req, res) => res.status(200).json(get()),
>     getTodo: (req, res) => res.status(200).json(getOne(req.params.id)),
>     addTodo: (req, res) => res.status(201).json(add(req.body.task)),
>     markTodoDone: (req, res) => res.status(201).json(done(req.params.id)),
>     markTodoNotDone: (req, res) => res.status(201).json(notDone(req.params.id)),
>     removeAllTodos: (req, res) => res.status(201).json(removeAll()),
>     removeTodo: (req, res) => res.status(201).json(remove(req.params.id) && get()),
>     updateTodo: (req, res) => res.status(201).json(update(req.params.id, req.body.task)),
>     errorHandler: (error, req, res) => res.status(400) && res.json(error)
> };
> ```
> 
> (which is also just translation of your current implementation)
> I think this way make it a little bit more cleaner, and you don't have to create variables will not be used in all calls.
> 
> > and i tested it and it is work perfectly
> 
> I'm pretty sure you did, but it is not only about working code, we are aiming at working and clean code.
> 
> Re: Error handling, please have a look at https://expressjs.com/en/guide/error-handling.html

Hello @eidosam ,
Thank you very much for your precious advises but i am a little bit lost and i did not get your point .
So I pushed already my old version of my homework .....please check.
best regards,

PR #385, opened 2019-07-02
Add and subtract from a state

ghifo:


week1/homework/src/server.js
```diff
@@ -9,12 +9,36 @@ function createServer(port) {
...
-    // TODO: Write your homework code here
+    console.log(request.method, request.url);
+
+    response.writeHead(200, { 'Content-Type': 'application/json' });
+    switch (request.url) {
+      case '/state':
+        response.write(JSON.stringify({ state }));
+        break;
+      case '/add':
+        state++;
```
gajduk: I appreciate that you separated app logic (addition) from technical details (stringify) 👍

week1/homework/src/server.js
```diff
@@ -9,12 +9,36 @@ function createServer(port) {
...
+        state--;
+        response.write(JSON.stringify({ state }));
+        break;
+      case '/reset':
+        state = 10;
+        response.write(JSON.stringify({ state }));
+        break;
+      default:
+        const error = 'Not found';
+        response.writeHead(404, { 'Content-Type': 'application/json' });
```
gajduk: 404 👍 

PRO TIP: you can also use `response.statusCode = 404;`

week1/homework/src/server.js
ghifo: Thanks!

week1/homework/src/server.js
ghifo:
```const error = 'Not found';
        response.writeHead(response.statusCode = 404, { 'Content-Type': 'application/json' });```
Do you mean this way?!

Comment (gajduk) [reply week1/homework/src/server.js]: I was thinking
```
response.statusCode = 404;
```
you can drop the `writeHead `line because  you already set the content type.
No big deal though

ghifo:
```default:
        const error = 'Not found';
        response.statusCode = 404;
        response.write(JSON.stringify({ error }));
        break;```

This way?

gajduk: Yes that is what I meant. At the very start you have
`response.writeHead(200, { 'Content-Type': 'application/json' });`
so there is no need to specify the content type two times.

ghifo: But this way I got `1 test failed`
with this comment after :
`  Expected status code to be 404

  Difference:

  - 200
  + 404`
because the code status was defined as `200` at the very beginning.

PR #386, opened 2019-07-02
Suheyb - Week1

suh3yb:


week1/homework/src/server.js
```diff
@@ -2,14 +2,38 @@
...
- * will fail.
- */
 function createServer(port) {
   let state = 10;
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    switch (request.url) {
+      case '/state':
+        response.setHeader('Content-Type', 'application/json');
```
gajduk: It is good practice to set Content-Type explicitly 👍

week1/homework/src/server.js
```diff
@@ -2,14 +2,38 @@
...
+        response.write(JSON.stringify({ state }));
+        break;
+      case '/add':
+        state++;
```
gajduk: Good idea to separate app logic `++` from technical details `stringify` 👍

week1/homework/src/server.js
```diff
@@ -2,14 +2,38 @@
...
+        response.write(JSON.stringify({ state }));
+        break;
+      case '/reset':
+        state = 10;
+        response.setHeader('Content-Type', 'application/json');
+        response.write(JSON.stringify({ state }));
+        break;
+      default:
+        response.statusCode = 404;
+        const error = new Error('Not found');
```
gajduk: using an Error without try catch makes very little sense. I would recommend using a string instead.
```
const error_message = 'Not found';
response.write(JSON.stringify({ error: error_message }));
```

week1/homework/src/server.js
```diff
@@ -2,14 +2,38 @@
...
+        response.setHeader('Content-Type', 'application/json');
+        response.write(JSON.stringify({ state }));
+        break;
+      case '/reset':
+        state = 10;
+        response.setHeader('Content-Type', 'application/json');
+        response.write(JSON.stringify({ state }));
+        break;
+      default:
+        response.statusCode = 404;
```
gajduk: 404 👍

suh3yb:
Hey @gajduk, thank you very much for your feedback. I changed `new Error` part of the code as you suggested.

gajduk:
> Hey @gajduk, thank you very much for your feedback. I changed `new Error` part of the code as you suggested.

Great.

I do not want you to necessarily code this but can you think of a way to write the code so that using an error is appropriate? Where would you throw the error? And where would the try catch be? If you want we can discuss this at week 3.

suh3yb:
> Great.
> 
> I do not want you to necessarily code this but can you think of a way to write the code so that using an error is appropriate? Where would you throw the error? And where would the try catch be? If you want we can discuss this at week 3.

Okay, I would like to discuss about it +1

PR #393, opened 2019-07-05
homework node wk1 Nadin

Nadoshka:


week1/homework/src/server.js
```diff
@@ -5,16 +5,44 @@ const http = require('http');
 /* `createServer` MUST return an instance of `http.Server` otherwise the tests
  * will fail.
  */
+function handleRequest(response, state) {
+  response.writeHead(200, { 'content-type': 'application/json' });
```
gajduk: It is great that you have put this logic in a separate function and reused it 👍

week1/homework/src/server.js
```diff
@@ -5,16 +5,44 @@ const http = require('http');
 /* `createServer` MUST return an instance of `http.Server` otherwise the tests
  * will fail.
  */
+function handleRequest(response, state) {
```
gajduk: 🤔 Is `handleRequest `a descriptive name for this function?

week1/homework/src/server.js
gajduk: It is good practice to set content-type explicitly 👍

week1/homework/src/server.js
```diff
@@ -5,16 +5,44 @@ const http = require('http');
...
+  response.writeHead(200, { 'content-type': 'application/json' });
+  response.write(JSON.stringify({ state }));
+}
+
+function handleError(response) {
```
gajduk: separate function for error's 👍

week1/homework/src/server.js
```diff
@@ -5,16 +5,44 @@ const http = require('http');
...
+        handleRequest(response, state);
+        break;
+      case '/reset':
+        state = 10;
+        handleRequest(response, state);
+        break;
+      default:
+        handleError(response);
+    }
+    response.end();
```
gajduk: response.end after break 👍

week1/homework/src/server.js
Nadoshka: Thanks! :)

week1/homework/src/server.js
Nadoshka: I Thought it was since it handles the request which comes from the client. :) So what do u think it could be a better name?

week1/homework/src/server.js
Nadoshka: Thanks a lot for your feedback! I appreciate it! :)

week1/homework/src/server.js
gajduk: You are right. Technically the function is handling a request. However, for a node application almost every function is related to handling requests :). The name should be more specific. 

Naming things is perhaps the hardest thing of all in programing. Here is a [nice post](https://dzone.com/articles/naming-conventions-from-uncle-bobs-clean-code-phil) based on a great book from Robert Martin aka Uncle Bob if you are interested.

PR #395, opened 2019-07-06
Hani's week 1 homework

hanialkudsi:


week1/homework/src/server.js
```diff
@@ -10,11 +10,38 @@ function createServer(port) {
...
+        response.write(JSON.stringify({ state }));
+        break;
+      case '/reset':
+        response.setHeader('content-Type', 'application/json');
+        state = 10;
+        response.write(JSON.stringify({ state }));
+        break;
+      default:
+        response.statusCode = 404;
+        let message = 'Not found';
```
eidosam: Better to use `const` instead of `let` for declaration if the variable is no re-assigned.

hanialkudsi: thanks for the nice comment's .I'll work on it next times

PR #398, opened 2019-07-09
hamza week2 homework

lizozo:
Hello;
I have revisited the homework since the first attempt was a little bit messy. You may ignore the first version and evaluate the revisited version.
Thank you


week2/homework/src/displayList.js
```diff
@@ -0,0 +1,12 @@
+const readFile = require('./readFile');
+
+async function displayList() {
+  const data = await readFile();
```
gajduk: Ooops
image

week2/homework/src/add.js
```diff
@@ -0,0 +1,13 @@
+function add(data, args) {
+  return new Promise(resolve => {
+    let keyOfNewItem = Object.keys(data).length + 1;
```
gajduk: This code looks like it  calculates the next available index i.e. key. But I do not understand why do you need the while loop then? Is one of them not working or is there another reason?

week2/homework/src/index.js
```diff
@@ -1,3 +1,76 @@
...
+const readFile = require('./readFile');
+const writeToFile = require('./writeToFile');
+const displayList = require('./displayList');
+const update = require('./update');
+const remove = require('./remove');
+const args = process.argv.slice(3);
+const command = process.argv[2];
+switch (command) {
+  case 'help':
+  default:
```
gajduk: TBH I did not know you can do this. 👍  for avoiding to duplicate the code

week2/homework/src/index.js
```diff
@@ -1,3 +1,76 @@
...
+    );
 
-// TODO: Write the homework code in this file
+    break;
+  case 'add':
+    (async () => {
+      try {
+        const data = await readFile();
+        const dataAddedArgs = await add(data, args);
+        await writeToFile(dataAddedArgs);
```
gajduk: use of await/async 👍

week2/homework/src/index.js
```diff
@@ -1,3 +1,76 @@
...
+    break;
+
+  case 'list':
+    displayList();
+    break;
+  case 'remove':
+    (async () => {
+      try {
+        const data = await readFile();
+        const removedData = await remove(data, args);
```
gajduk: I really like that you have separated the whole file logic from the rest of the app logic 👍

week2/homework/src/index.js
```diff
@@ -1,3 +1,76 @@
...
+        await writeToFile(removedData);
+        displayList();
+      } catch (error) {
+        console.error;
```
gajduk: I think you meant `console.error(error)`

week2/homework/src/readFile.js
```diff
@@ -0,0 +1,15 @@
+const fs = require('fs');
+
+function readFile() {
+  return new Promise((resolve, reject) => {
```
gajduk: proper use of promises

week2/homework/src/add.js
lizozo: Hello @gajduk !! thanks for the feedback and any i really appreciate your time and effort.

 it calculates the available index based on the length of the object. suppose that i have an object with 3 properties. 
( { "1" : "goto market" , "2" : "visit mom", "3" : "see the dentist"} ). so the next property will be length of my object (3) +1 =4. But suppose that i removed 2 from the list. then my object is:
( { "1" : "goto market" , "3" : "see the dentist"} ) So when i add new item, the key will be length of my object(2) +1 =3. But 3 is already in the list("3" : "see the dentist"). So i check if the new key is in the list with: .hasOwnProperty(key). if the newly created key exists in the list, it is incremented in the while loop until it doesnt conflict with any of the key in the object.

By the way my code works fine in my computer and does not return any error. sorry for the inconvenience. I checked the displayList method and I think the error is caused by not handling rejected state of the promise. so i added try-catch block. I am not able to test it, becuz i dont get any error. But anyway I will make a new pull request. I hope it works fine.
thank you very much again
Kind regards

week2/homework/src/index.js
```diff
@@ -49,7 +49,7 @@ switch (command) {
     break;
 
   case 'list':
-    displayList();
+    displayList().catch(error => console.log(error));
```
gajduk: At least know the error is handled 👍

PR #400, opened 2019-07-10
week2

zehrakopar:


week2/homework/src/add.js
```diff
@@ -0,0 +1,10 @@
+const fs = require('fs');
+
+function add(todoText) {
+  fs.appendFile('./src/data.json', todoText + '\n', error => {
+    if (error) throw error;
```
gajduk: rethrow error 👍

week2/homework/src/index.js
```diff
@@ -1,3 +1,43 @@
 'use strict';
 
-// TODO: Write the homework code in this file
+const add = require('./add');
```
gajduk: separate files per command 👍

week2/homework/src/index.js
```diff
@@ -1,3 +1,43 @@
...
+  case 'reset':
+    reset();
+    break;
+
+  case 'update':
+    update(index, updatedText);
+    break;
+
+  default:
+    console.log(`${command} is not a valid command`);
```
gajduk: very clean :+1

week2/homework/src/list.js
```diff
@@ -0,0 +1,14 @@
+const fs = require('fs');
+
+function list() {
+  fs.readFile('./src/data.json', 'utf8', (error, data) => {
+    if (error) {
+      console.log(error);
```
gajduk: Why did you decide to log the error instead of rethrowing it like in some `add`?

week2/homework/src/update.js
```diff
@@ -0,0 +1,21 @@
+const fs = require('fs');
+
+function update(index, updatedText) {
+  fs.readFile('./src/data.json', 'utf8', (error, data) => {
```
gajduk: Is `.json` the right extension?

week2/homework/src/list.js
zehrakopar: i don't know. you are are it is better to use throw error.

week2/homework/src/update.js
zehrakopar: in the assignment guadeline it is advised...

gajduk: That is true, but you are using plain text to store the data. Calling a file data.json does not magically make it a json file. In fact it is misleading as people will assume that the data in the file is stored in a json format and will try to parse it. If you are storing data as plain text the file should be data.txt

PR #408, opened 2019-07-17
Suheyb - Week3

suh3yb:


week3/homework/src/actions.js
```diff
@@ -0,0 +1,158 @@
...
+        return listAndIndex === undefined
+          ? reject(new ErrorWithStatus(`No to-do is found with id: ${id}`, 404))
+          : resolve(listAndIndex);
+      })
+      .catch(error => reject(error));
+  });
+}
+
+function isToDoValid(request) {
+  return new Promise((resolve, reject) => {
```
gajduk: does this really need to be a promise?

week3/homework/src/index.js
```diff
@@ -1,3 +1,74 @@
...
+const app = express();
+
+const PORT = 3000;
+
+app.use(express.json());
+
+app.get('/', (request, response) => {
+  readAndParse('help')
+    .then(help => response.json(help))
+    .catch(error => response.status(error.statusCode || 500).json(handleError(error)));
```
gajduk: This part seems to be repeating in every call. There is a way to eliminate the duplication:
https://expressjs.com/en/guide/error-handling.html
This is not part of the homework but might be interesting for you to know.

week3/homework/src/actions.js
```diff
@@ -0,0 +1,158 @@
...
+
+class ErrorWithStatus extends Error {
+  constructor(errorMessage, statusCode) {
+    super(errorMessage);
+    this.name = 'ErrorWithStatus';
+    this.statusCode = statusCode;
+  }
+}
+
+function findToDoWithId(id) {
```
gajduk: some of these function do not seem to be actions. Perhaps they deserve to be in their own file?
This file is already 150 lines long :(

suh3yb:
Thank you very much for your feedback @gajduk. 
I will make the necessary changes and try to convert the code into async/await.

week3/homework/src/actionsAndStaff/writeFindCheckError.js
```diff
@@ -0,0 +1,62 @@
...
+    super(errorMessage);
+    this.name = 'ErrorWithStatus';
+    this.statusCode = statusCode;
+  }
+}
+
+async function write(jsonData) {
+  try {
+    await fs.writeFile(LIST_FILE_PATH, JSON.stringify(jsonData, null, 2), error => {
+      if (error) throw error;
```
gajduk: I have not tested this but I think it will not work. You need to promisify writeFile first
```
await new Promise((resolve,reject) => fs.writeFile("foo.txt", "example", error => {
       if ( error) reject(error);
       resolve();
}
```
Or use https://www.npmjs.com/package/promise-fs

week3/homework/src/index.js
```diff
@@ -17,58 +17,76 @@ const PORT = 3000;
 
 app.use(express.json());
 
-app.get('/', (request, response) => {
-  readAndParse('help')
-    .then(help => response.json(help))
-    .catch(error => response.status(error.statusCode || 500).json(handleError(error)));
+app.get('/', async (request, response) => {
```
gajduk: there is a trick to avoid writing try catch all the time:
```
/**
* function that takes another (async) function as a parameter
* then wraps it in a try catch
* obviously the name is arbitrary
* we only need to define this once
**/
const wrap = function(fn) { 
    return async function(req, res, next) {
             try { 
                 res.json(await fn(req, res, next));
            }
            catch(error) {
                 response.status(error.statusCode || 500).json(handleError(error);
            }
     };
}

//then when defining the handlers we use the function to wrap them
// ta dam one liner
app.get('/', wrap(async (request, response) => await readAndParse('help') );
```

week3/homework/src/actionsAndStaff/markTodo.js
```diff
@@ -0,0 +1,15 @@
...
+
+const { findToDoWithId, write } = require('./writeFindCheckError');
+
+const mark = async (id, requestMethod) => {
+  try {
+    const { toDoList, index } = await findToDoWithId(id);
+    toDoList.todos[index].done = requestMethod === 'DELETE' ? false : true;
+    return write(toDoList);
+  } catch (error) {
+    throw error;
```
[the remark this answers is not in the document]
suh3yb: Does this mean I could get rid of `try/catch` here even if I use `async/await`?

week3/homework/src/actionsAndStaff/writeFindCheckError.js
suh3yb: I have tested as a whole and everything was working. Perhaps something that I could not think of while testing would cause a problem here. I will use `promisify` anyway, so that the code seems more consistent since I use promise for `fs.readFile`.

week3/homework/src/actionsAndStaff/markTodo.js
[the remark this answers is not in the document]
gajduk: yes.

suh3yb:
Thank you very much for your time and feedback @gajduk.
I got rid of some `try/catch`s and used the function as you suggested.
Best regards.

PR #410, opened 2019-07-17
Express app

ghifo:


week3/homework/src/createTodo.js
```diff
@@ -0,0 +1,24 @@
...
+async function createTodo(description) {
+  const todos = await readTodos();
+  const todo = {
+    id: uuid(),
+    done: false,
+    description,
+  };
+  todos.push(todo);
+
+  return fs.writeFile(fileName, JSON.stringify(todos, null, 2), error => {
```
gajduk: Is this the correct way to convert from a callback to async (Promise)?

week3/homework/src/index.js
```diff
@@ -1,3 +1,71 @@
...
+const express = require('express');
+
+const readTodos = require('./readTodos');
+const createTodo = require('./createTodo');
+const deleteTodo = require('./deleteTodo');
+const updateTodo = require('./updateTodo');
+const readTodo = require('./readTodo');
+const clearTodos = require('./clearTodos');
+const markAsDone = require('./markAsDone');
+const markAsNotDone = require('./markAsNotDone');
```
gajduk: separate files 👍

week3/homework/src/index.js
```diff
@@ -1,3 +1,71 @@
...
+
+const app = express();
+
+app.use(express.json());
+
+try {
+  app.get('/todos', (request, response) => {
+    readTodos().then(todos => response.status(200).json(todos));
```
gajduk: I really like that you used `async` for the `readTodos` function. 
But why do you use then here. Wasn't the whole point of using `async` to use `await` instead of `then`?
How can you write this code so that is uses `await` instead of `then`?

week3/homework/src/index.js
```diff
@@ -1,3 +1,71 @@
...
+  app.get('/todos/:id', (request, response) => {
+    const id = request.params.id;
+    readTodo(id).then(todo => {
+      response.status(200).json(todo);
+    });
+  });
+
+  app.post('/todos', (request, response) => {
+    const addTodo = request.body;
+    const description = addTodo.description;
```
gajduk: What happens if there is no description? Some validation would be nice

week3/homework/src/index.js
ghifo: I got this error message when I used  await instead of then.

` const todos = await readTodos()`

PR #411, opened 2019-07-18
7alip | Week3

7alip:


week3/homework/src/Todo.js
```diff
@@ -0,0 +1,112 @@
...
+const uuidv4 = require('uuid/v4');
+const util = require('util');
+const fs = require('fs');
+
+const readPromise = util.promisify(fs.readFile);
+const writePromise = util.promisify(fs.writeFile);
+
+const DEFAULT_ENCODING = 'utf8';
+
+class Todo {
```
gajduk: Interesting approach 👍

week3/homework/src/Todo.js
```diff
@@ -0,0 +1,112 @@
...
+    const todos = await this.read();
+    const todo = await this.read(id);
+
+    todo.text = text;
+    await this._save(todos);
+
+    return todo;
+  }
+
+  async delete(id) {
```
gajduk: great job on using async/await 👍

week3/homework/src/index.js
```diff
@@ -1,3 +1,26 @@
...
+
+const express = require('express');
+const app = express();
+
+app.use(express.json());
+
+const Todo = require('./Todo');
+const todo = new Todo(FILE_PATH);
+
+app.get(TODO_URL, readTodos.bind(null, todo));
```
gajduk: very tidy 👍

week3/homework/src/todos.json
```diff
@@ -0,0 +1,17 @@
+[
```
gajduk: What is a json file with data doing in a `src` directory?

7alip: I forgot to delete json file after I test post requests.

7alip:
Hi @gajduk
Thanks for feedback and nice comments. 

I didn't test my code after my last change. When the file updated I saved the old data but not the data which is changed. I noticed my problem here:
```javascript
    const todos = await this.read();
    const todo = await this.read(id); 
    // const todo = todos.find(t => t.id === id); // => It should be just like in the previous commit
```
That's why ```todos``` array was not effected when I update ```todo``` object. 

I also made some changes about catching errors when we try to read non-exist json file.

PR #414, opened 2019-07-19
alhommada week3 node.js

aalhommada:


week3/homework/src/index.js
```diff
@@ -3,80 +3,81 @@
...
-    id: todos.length + 1,
-    name: req.body.name,
+  const todoItem = {
+    id: todo.length + 1,
+    name: req.body.description,
     done: false,
   };
-  todos.push(todo);
-  res.send(todos);
+  todo.push(todoItem);
```
gajduk: This is not saved. You have to explicitly write this to the file system.

week3/homework/src/index.js
```diff
@@ -3,80 +3,81 @@
 const express = require('express');
 const Joi = require('joi');
 const app = express();
-const todos = require('./todos');
+const validateTodos = require('./validateTodos');
+const todo = require('./todo.json');
```
gajduk: Please dont read json files using require. It is a really bad practice. Use `fs.readFile` like we did in the lectures.

week3/homework/src/index.js
aalhommada: Thanks i will update it

week3/homework/src/index.js
aalhommada: Thanks . I will modify it

week3/homework/src/index.js
```diff
@@ -1,84 +1,100 @@
...
+    } else {
+      const todo = JSON.parse(contents);
+      const todoItem = {
+        id: todo.length + 1,
+        description: req.body.description,
+        done: false,
+      };
+      todo.push(todoItem);
+      const newTodo = JSON.stringify(todo);
+      fs.writeFile('/todo.json', newTodo, error => {
```
gajduk: exactly like this 👍

aalhommada: Thanks a lot @gajduk ,  I learn alot during your review.

PR #417, opened 2019-07-20
week3

zehrakopar:


week3/homework/src/index.js
```diff
@@ -1,3 +1,55 @@
...
+const remove = require('./remove');
+const reset = require('./reset');
+const readTodo = require('./readTodo');
+const markAsDone = require('./markAsDone');
+const markAsNotDone = require('./markAsNotDone');
+
+app.use(express.json());
+
+try {
+  app.get('/todos/:id', (request, response) => {
```
gajduk: there should  also be a get todos that returns a list of all todos

week3/homework/src/index.js
```diff
@@ -1,3 +1,55 @@
...
+    readTodo(request.params.id);
+    response.status(200).json({ message: `todo with the id of ${request.params.id} is requested` });
+  });
+
+  app.post('todos/:id/:text', (request, response) => {
```
gajduk: this is not the correct format for creating todos. Please review the lecture video. You need to use the body of the requests

week3/homework/src/index.js
```diff
@@ -1,3 +1,55 @@
...
+    add(request.params.id, request.params.text);
+    response.status(201).json({ message: `todo with the id of ${request.params.id} is added` });
+  });
+
+  app.put('todos/:id/:text', (request, response) => {
```
gajduk: same here

fedefu:
Thanks for the feedback @gajduk! @zehrakopar can you please implement the feedback?

week3/homework/src/index.js
zehrakopar: i ve made this changes

gajduk:
Good job.

PR #418, opened 2019-07-20
Veli | Node.js-Week3

veliataseven:


week3/homework/src/functions.js
```diff
@@ -0,0 +1,196 @@
...
+    return { error: `${error.message}` };
+  }
+}
+
+function isInputValid(request) {
+  try {
+    if (request.headers['content-type'] === 'application/json') {
+      if (request.body.hasOwnProperty('todo')) {
+        if (request.body.todo.hasOwnProperty('description')) {
+          if (
```
gajduk: nice to add validations 👍

week3/homework/src/functions.js
```diff
@@ -0,0 +1,196 @@
...
+    if (isFileExist()) {
+      if (isInputValid(request)) {
+        request.body.todo.id = uuidv4();
+        request.body.todo.done = false;
+        const obj = JSON.parse(fs.readFileSync(fileName, 'utf8'));
+        obj.todoArray.push(request.body);
+        fs.writeFileSync(fileName, JSON.stringify(obj));
+        return obj;
+      }
+      return { error: 'content-type is not application/json or todo text is not valid' };
```
gajduk: which one is it? content type is not application/json or the todo text is not valid?
If only one could know...
Is there a simple way to propagate this information up the call stack 🤔  ?

week3/homework/src/functions.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+}
+
+function getToDoById(id) {
+  try {
+    if (isFileExist()) {
+      const obj = JSON.parse(fs.readFileSync(fileName, 'utf8'));
+      const result = obj.todoArray.filter(element => {
+        if (element.todo.id == id) {
+          return element;
```
gajduk: I first thought, that this is a mistake. It turns out to work properly, but can you explain why?

week3/homework/src/functions.js
```diff
@@ -0,0 +1,196 @@
...
+  } catch (error) {
+    return { error: `${error.message}` };
+  }
+}
+
+function markToDo(id, method) {
+  try {
+    if (isFileExist()) {
+      const obj = JSON.parse(fs.readFileSync(fileName, 'utf8'));
+      const result = obj.todoArray.filter(element => {
```
gajduk: why not use forEach ?

week3/homework/src/index.js
```diff
@@ -1,3 +1,48 @@
...
+const express = require('express');
+const app = express();
+const port = 3000;
+
+const functions = require('./functions');
+
+app.use(express.json());
+app.listen(port, () => console.log(`Example app listening on port ${port}!`));
+
+app.get('/', (request, response) => {
```
gajduk: this file is very clean and tidy

week3/homework/src/functions.js
veliataseven: Filter let you provide a callback for every element and returns a filtered array.
The main difference between forEach and filter is that forEach just loop over the array and executes the callback but filter executes the callback and check its return value. If the value is true element remains in the resulting array but if the return value is false the element will be removed for the resulting array..

After filtered the array I get an new array that provides some conditions . And then I am checking the if new array has an item or not.. If new array's length == 0, new array has no item.. I mean filtered array return empty array..

week3/homework/src/functions.js
veliataseven: What is the mistake that You thought? maybe I can explain it after that

week3/homework/src/functions.js
veliataseven: Yes you re right.. Actually I knew that I must explain easily with just some checking some condition(if -else etc) it that which one is it? content type is not application/json or the todo text is not valid?..  that was my lazy..

PR #419, opened 2019-07-23
Week3

hanialkudsi:


week3/homework/src/write.js
```diff
@@ -0,0 +1,16 @@
+const fs = require('fs');
+
+function write(req, res, toBeWritten) {
+  return new Promise((resolve, reject) => {
```
gajduk: nice that you promisified fs.writeFile 👍

week3/homework/src/markDone.js
```diff
@@ -0,0 +1,27 @@
+const write = require('./write');
+const read = require('./read');
+
+function markDone(req, res, toDelete) {
+  read().then(parsedData => {
+    parsedData.map((toDo, index) => {
```
gajduk: why did you decide to use map here instead of find like in the other functions?
filter, find, map etc all take a function as argument which should be a _pure_ function https://medium.com/javascript-scene/master-the-javascript-interview-what-is-a-pure-function-d1c076bec976
In other words the function should not have side effects (such as `todo.done = true`)

gajduk: A much nicer way to write would be:

```
read().then(parsedData => {
    const todoItem = parsedData.find(todo => todo.id == req.query.id);
    todoItem.done = ! todoItem.done;
   ....
```

week3/homework/src/index.js
```diff
@@ -1,3 +1,50 @@
...
+const deleteAll = require('./deleteAll');
+const deleteOne = require('./deleteOne');
+const addOne = require('./addToDo');
+const markDone = require('./markDone');
+const editToDo = require('./editToDo');
+
+app.use(express.json());
+app.use(bodyParser.json());
+
+app.post('/addTodo', (req, res) => {
```
gajduk: In a restful API is the url is a noun and the http method is the verb
todo = noun
post = verb = create

The idea is precisely to avoid urls such as addTodo/createTodo/newTodod (mainly to avoid confusion, as you can see there are many way how to name a thing as simple as creating a new todo)

week3/homework/src/index.js
```diff
@@ -1,3 +1,50 @@
...
+  const toWrite = req.body.todo.description;
+  addOne(req, res, toWrite);
+});
+
+app.get('/todos', (req, res) => {
```
gajduk: the noun is _todo_ and should be consistent

week3/homework/src/index.js
```diff
@@ -1,3 +1,50 @@
...
+  getList(req, res);
+});
+
+app.get('/todo/?', (req, res) => {
```
gajduk: this is not the preferred REST way. Use `/todo/:id` instead

week3/homework/src/index.js
```diff
@@ -1,3 +1,50 @@
...
+app.delete('/todos', (req, res) => {
+  deleteAll(req, res);
+});
+
+app.put('/todo/?', (req, res) => {
+  deleteOne(req, res, req.query.id);
+});
+
+app.put('/todoDone/?', (req, res) => {
+  markDone(req, res, req.query.id);
```
gajduk: this file reads very cleanly because all the logic is in other files 👍

fedefu:
thanks for the feedback @gajduk. Can you please update your code @hanialkudsi

week3/homework/src/index.js
hanialkudsi: of course I update this

week3/homework/src/index.js
hanialkudsi: i'll work on that hopefully it's better now

week3/homework/src/markDone.js
hanialkudsi: yes that's true it's even shorter thanks a lot

week3/homework/src/write.js
hanialkudsi: thanks man

fedefu:
Hi there @Meno-101, @hanialkudsi and @gajduk. I see the feedback given to this PR but not the implementations submitted. This needs to be approved or redone @hanialkudsi. Please let me know how you proceed.

gajduk:
@hanialkudsi the most important feedback to implement is https://github.com/HackYourFuture/Node.js/pull/419#discussion_r307914414 and
https://github.com/HackYourFuture/Node.js/pull/419#discussion_r307914676

PR #422, opened 2019-08-08
Suggestion - remove part 1

gajduk:
If I look at the homework for week1 in masters it is only about setting up a simple server with few endpoints. Maybe it is enough for them to set up the project, get accustomed to npm and node.


gajduk:
I made a new suggestion.

PR #424, opened 2019-08-13
added week1 homework- Salih S

salih18:
The PR also includes extra homework.

### DIFF
--- a/week1/homework/src/server.js
+++ b/week1/homework/src/server.js
@@ -2,19 +2,23 @@
 
 const http = require('http');
 
-/* `createServer` MUST return an instance of `http.Server` otherwise the tests
- * will fail.
- */
+const getResponse = (responseStatus, response, responseText) => {
+  response.writeHead(responseStatus, { 'Content-Type': 'application/json' });
+  response.end(JSON.stringify(responseText));
+};
+
 function createServer(port) {
   let state = 10;
-
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    if (request.url === '/state') getResponse(200, response, { state: state });
+    else if (request.url === '/add') getResponse(200, response, { state: ++state });
+    else if (request.url === '/subtract') getResponse(200, response, { state: --state });
+    else if (request.url === '/reset') getResponse(200, response, { state: (state = 10) });
+    else getResponse(404, response, { error: 'Not found' });
   });
-
   return server;
 }
 
 module.exports = {
-  createServer
+  createServer,
 };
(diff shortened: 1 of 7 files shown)


week1/homework/src/server.js
```diff
@@ -2,16 +2,20 @@
...
+
 function createServer(port) {
   let state = 10;
-
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    if (request.url === '/state') getResponse(200, response, { state: state });
+    else if (request.url === '/add') getResponse(200, response, { state: ++state });
+    else if (request.url === '/subtract') getResponse(200, response, { state: --state });
+    else if (request.url === '/reset') getResponse(200, response, { state: (state = 10) });
```
Meno-101: In a case of using a single variable with many expected values, we could use a `switch` statement instead.

It will be a lot more legible for readers and will tell what it may contain of case uses.
And `(state = 10)` will be on it's own line as it's reassigned either ways, but it will be telling it's been reassigned :)

week1/homework/src/server.js
```diff
@@ -2,16 +2,20 @@
 
 const http = require('http');
 
-/* `createServer` MUST return an instance of `http.Server` otherwise the tests
- * will fail.
- */
+const getResponse = (responseStatus, response, responseObj) => {
```
Meno-101: it's probably a valid name from your point of view, but from mine I would prefer `sendResponse` since it's sending it, and no big deal 😄

Meno-101:
And I had a look at your extra homework it looks very good to me.

week1/homework/src/server.js
salih18: Hi Jawhar, thank you so much for your review. I think, the reason why I prefer not to use switch-case is  while we were in the first JavaScript lessons our teacher imposed us 'switch-case is not preferred when you write codes in the work.' :)) Actually I thought to use ternary instead of if else :)

week1/homework/src/server.js
salih18: Sometimes while writing the code, the more difficult one is to pick correct function names, generally I am not lucky about this. :)) I thought like this while choosing the name.  'sending request' and 'getting response'

week1/homework/src/server.js
Meno-101: Yes that's true, but it's always if you have a variable definition and it's value on of a set of values, I will go for the switch statement.

```javascirpt
if (variable === '/state') {
} else if (variable === '/other-state') {
} else {
}
```

```javascirpt
switch (variable) {
  case '/state':
  case '/other-state':
  default :
}
```

PR #425, opened 2019-08-13
Submit week1 homework - Oguzhan

oguzhanturgut:
Extra homework is in `extra-homework` folder


week1/homework/src/server.js
```diff
@@ -1,15 +1,43 @@
...
 
 const http = require('http');
-
+const State = require('./state');
 /* `createServer` MUST return an instance of `http.Server` otherwise the tests
  * will fail.
  */
 function createServer(port) {
-  let state = 10;
+  const state = new State();
```
Meno-101: I liked this constructor.

week1/homework/src/server.js
```diff
@@ -1,15 +1,43 @@
...
+  function sendState(response) {
+    response.writeHead(200, { 'Content-Type': 'application/json' });
+    response.end(JSON.stringify(state));
```
Meno-101: You are duplicating some of the same code for the same use,
this could be better in a matter of shortness, what about:
```javascript
// now this can be used in case of success or failer statuses, and you will have less lines
function sendJson(status, response, content) {
  response.writeHead(status, { 'Content-Type': 'application/json' });
  response.end(JSON.stringify(content));
}
```

week1/extra-homework/sendJS.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const fs = require('fs');
+function sendJS(res) {
+  res.writeHead(200, { 'Content-Type': 'ext/javascript' });
```
Meno-101:
```suggestion
  res.writeHead(200, { 'Content-Type': 'text/javascript' });
```

week1/extra-homework/server.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+const http = require('http');
+const sendIndex = require('./sendIndex');
+const sendJS = require('./sendJS');
+const sendImage = require('./sendImage');
+const sendStyle = require('./sendStyle');
+
+function createServer(port) {
+  function handleRequest(req, res) {
```
Meno-101: Since it's mostly we are handling the response to be served, or in other words, we are creating a server.
is handle request a descriptive name?

week1/extra-homework/package.json
```diff
@@ -0,0 +1,11 @@
+{
```
Meno-101: No need for this `package.json` file, since we haven't made a have a library and it's not requires any dependencies.

oguzhanturgut: I made this `package.json` just for keeping the extra-homework as a different app with app name and description. You're right that i didn't add any dependency to this app.

week1/extra-homework/server.js
oguzhanturgut: You're totally right. Since we are at the server side, we are getting the request from the client and handling(dealing with) the response because `request` is read-only for us whereas `response` is writable. On the other hand documentation calls this function as `requestListener`. I am a bit confused with this callback and not very good at naming functions or variables 😁
Changed to `handleResponse`

week1/extra-homework/sendJS.js
oguzhanturgut: 😱Sneaky typo. Fixed it. Thanks for the tip

week1/homework/src/server.js
oguzhanturgut: At first i deliberately isolated the error and state sending functions due to 'separation of concerns'. But as you mentioned with refactoring they are still very close concerns. No need to separate.
Refactored into  `sendJson`
Thanks for the tip

week1/homework/src/server.js
oguzhanturgut: Thank you

oguzhanturgut:
> Hello @oguzhanturgut,
> 
> Amazing,
> 
> **no mistakes, clean code.** And you have mad the bonus ❤️
> I just have left for you two comments, please have a look and tell me your thoughts, and then I approve 🎉

Hello @Meno-101 
Thanks for your supportive and wonderful comments. I'd appreciate your time and effort for this feedback. Have a nice and sunny day ☀️

PR #426, opened 2019-08-14
Adham node js homework week1

Adham-Elias-Botours:
This is my NodeJs homewoek week1 .
It was very difficult to understand what the question is about and what I have to do!
Also understanding the relation between the 2 files  `index.js` and `server.js` was difficult.

* I do not understand why I got ESLint error with the (if, else if, else) version of my code. 
Eslint insisted to change the format (indent, and space) depending on its default setting and then it caused error by it self. I do not know yet how to control that. But on the other hand the version passed the test.

In this current version I used `switch`. It doesn't cause any ESLint error.

I hope this work to be good start for me with NodeJS.
Thanks in advanced.

### DIFF
--- a/week1/homework/src/server.js
+++ b/week1/homework/src/server.js
@@ -10,7 +10,6 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
-    console.log(request.url);
     const giveResult = (statusCode, object) => {
       response.writeHead(statusCode, { 'Content-Type': 'application/json' });
       response.write(JSON.stringify(object));
--- a/week1/homework/src/server.js
+++ b/week1/homework/src/server.js
@@ -17,31 +17,21 @@ function createServer(port) {
     };
     if (request.url === `/state`) {
       giveResult(200, { state });
-      // response.writeHead(200, { 'Content-Type': 'application/json' });
-      // response.write(JSON.stringify({ state }));
     }
  else if (request.url === `/add`) {
       state++;
       giveResult(200, { state });
-      // response.writeHead(200, { 'Content-Type': 'application/json' });
-      // response.write(JSON.stringify({ state }));
     }
  else if (request.url === `/subtract`) {
       state--;
       giveResult(200, { state });
-      // response.writeHead(200, { 'Content-Type': 'application/json' });
-      // response.write(JSON.stringify({ state }));
     }
  else if (request.url === `/reset`) {
       state = 10;
       giveResult(200, { state });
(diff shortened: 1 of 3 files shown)


week1/homework/src/server.js
```diff
@@ -10,6 +10,33 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    const giveResult = (statusCode, object) => {
+      response.writeHead(statusCode, { 'Content-Type': 'application/json' });
+      response.write(JSON.stringify(object));
+    };
+
+    switch (request.url) {
```
Meno-101: As your comment says,
> In this current version I used switch. It doesn't cause any ESLint error.

That's right! looking back to [google's style guide], it's better to use in this manner, I would say it's even better to use it when ever you have a code similar to this kind of thing **(not in the guides)**, when you have a variable definition and it's value one of a set of a values, I will go for the switch statement.

And that's why `Eslint` was complaining 🙂 

```javascript
if (variable === '/state') {
} else if (variable === '/other-state') {
} else {
}
```

```javascript
switch (variable) {
  case '/state':
  case '/other-state':
  default :
}
```

[google's style guide]: https://google.github.io/styleguide/jsguide.html#features-switch-statements

week1/homework/src/server.js
```diff
@@ -10,6 +10,33 @@ function createServer(port) {
...
+      default: giveResult(404, { error: `Not found` });
```
Meno-101: for a best practice and more legibility I would go for keeping this `default` case at the end of my code.

week1/homework/src/server.js
```diff
@@ -10,6 +10,33 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    const giveResult = (statusCode, object) => {
```
Meno-101: Nice DRY one!

week1/homework/src/server.js
Adham-Elias-Botours: I have never heard about: google's style guide! 
Thanks, it is very useful. The layout and how the information  are displayed , make it very  comfortable and clear.  I like it.

week1/homework/src/server.js
Adham-Elias-Botours: My code is a reflection of how I thought while I writing it. I started with the default situation, then started to add  the others possibility one buy one.  after finishing I searched again to know if it is normaal to keep the default on the top. And the answer was: yes it is possible,  so I did not change it. 
But on the other hand I agree with what you mentioned.

PR #428, opened 2019-08-15
add homework week1

musafir928:


week1/homework/src/server.js
```diff
@@ -10,6 +10,53 @@ function createServer(port) {
...
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+
+    switch (request.url) {
+      case '/add':
+        response.writeHead(200, {
+          'Content-Type': 'application/json'
+        });
+
+        response.end(JSON.stringify({
```
Meno-101: Since we are duplicating all of this code from line 16 to line 22 with some tiny changes each time, perhaps a DRY function will make look a bit more shorter?

And in that case you will find it even more readable almost as:
```javascript
      case '/add':
        // Short code.
        break;
      case '/state':
        // Short code.
        break;
      case '/subtract':
        // Short code.
        break;
      case '/reset':
        // Short code.
        break;
      default:
        // Short code.
    }
```

week1/homework/src/server.js
```diff
@@ -10,6 +10,53 @@ function createServer(port) {
...
+          state: ++state
+        }, null, 2));
```
Meno-101: From all the `case`s you have,
you have some extras no need for them unless you are using it for let's say writing to a file or view the object somewhere you want it to be beautified.

```suggestion
        }));
```

week1/homework/src/server.js
musafir928: Thank you for your time and feedback.

week1/homework/src/server.js
musafir928: I have fixed it in writeResponse function.

musafir928:
> well done!
> Thanks for the changes :)

musafir928:
@Meno-101  Thanks a lot for your review and comment!

PR #430, opened 2019-08-16
nodejs-w1-homework-Shyar

ShyarK:


week1/homework/src/fileContent.js
```diff
@@ -0,0 +1,13 @@
+function State(num) {
```
Meno-101: Very nice implementation for a constructor 🎉

week1/homework/src/helper.js
```diff
@@ -0,0 +1,12 @@
+class Util {
+  static handleError(response) {
+    response.writeHead(404, { 'Content-Type': 'application/json' });
+    response.end(JSON.stringify({ error: 'Not found' }));
+  }
+  static sendFile(response, state) {
+    response.writeHead(200, { 'content-type': 'application/json' });
+    response.end(JSON.stringify(state));
```
Meno-101: As line 8 this method is probably ending the response, or in other words it ends the process of the response and send it back to the client.

Straight to the point is this method have the right name of it's functionality? and `fileContent.js` file too?

week1/homework/src/server.js
```diff
@@ -1,15 +1,32 @@
...
  * will fail.
  */
-function createServer(port) {
-  let state = 10;
 
+function createServer(port) {
+  const state = new State(10);
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    if (request.url === '/state') {
```
Meno-101: For a code style purpose, I would go for using a `switch` statement here instead of the `if` conditional, for readability and avoiding any kind of an errors, such as trying to send the response two times, first the actual response and the other is when you are handling the error:

```javascript
if (variable === '/state') {
}
if (variable === '/other-state') {
}
// in `express` this will throw an error, in normal `http` it will throw too if it sets the Headers again after they are set.
Util.handleError(response);
```

```javascript
switch (variable) {
  case '/state':
  case '/other-state':
  default :
}
```

PR #431, opened 2019-08-16
submission Node Js Hw Week1

Mehmet-Ergorgec:


week1/homework/src/server.js
```diff
@@ -5,11 +5,41 @@ const http = require('http');
...
+  }
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    switch (request.url) {
+      case '/state':
+        handleRequest(response, 'application/json');
+        break;
+      case '/add':
+        state++;
```
gajduk: Great to see you separated the logic of the application i.e increment state from the technical details such as setting head ers and status codes.

week1/homework/src/server.js
```diff
@@ -5,11 +5,41 @@ const http = require('http');
 /* `createServer` MUST return an instance of `http.Server` otherwise the tests
  * will fail.
  */
+
 function createServer(port) {
   let state = 10;
+  function handleRequest(response, fileType) {
+    response.writeHead(200, { 'Content-Type': fileType });
+    response.write(JSON.stringify({ state }));
```
gajduk: I am really happy that you isolated this in a separate function that you are reusing. A small thing that bothers me is that the response body is always` JSON.stringify(state)` i.e. the response content type is always JSON. Thats ok, but if you already know that is the case, does it make sense to have the content-type as a parameter?

week1/homework/src/server.js
```diff
@@ -5,11 +5,41 @@ const http = require('http');
 /* `createServer` MUST return an instance of `http.Server` otherwise the tests
  * will fail.
  */
+
 function createServer(port) {
   let state = 10;
+  function handleRequest(response, fileType) {
```
gajduk: In the context of a server every code is in a way handling a request. So handleRequest is not a great name for a function. Can you think of a better name?

week1/homework/src/server.js
Mehmet-Ergorgec: I thought how I can make it reusable. Yes You are right. In this example always JSON.

week1/homework/src/server.js
Mehmet-Ergorgec: Maybe handleResponse or sendResponse?

PR #433, opened 2019-08-16
homework week1 submission

hnnhack:


week1/homework/src/server.js
```diff
@@ -1,20 +1,42 @@
...
+          postData(200, response, { state: state });
+          break;
+        case '/add':
+          postData(200, response, { state: ++state });
+          break;
+        case '/subtract':
+          postData(200, response, { state: --state });
+          break;
+        case '/reset':
+          postData(200, response, { state: (state = 10) });
```
gajduk: interesting approach. Keep in mind that in other languages assignment operation does not always return the same result. For example in c I believe the result of an assignment is always 1.  Sometimes using such clever tricks can fire back and lead to a very painful debugging session. So best approach would be to keep things explicit even if that means writing more lines of code.

week1/homework/src/server.js
```diff
@@ -1,20 +1,42 @@
...
-const http = require('http');
+{
+  const http = require('http');
 
-/* `createServer` MUST return an instance of `http.Server` otherwise the tests
- * will fail.
- */
-function createServer(port) {
-  let state = 10;
+  const postData = (status, response, data) => {
```
gajduk: nice that you separated this in a function 👍

hnnhack: Hi @gajduk !
Thank you very much for your efforts to review and making a note.

PR #435, opened 2019-08-17
finished node homework week01

Laurelinduilwen:
for the bonus homework regarding the file server I have created a new repo at: 

https://github.com/Laurelinduilwen/basic-node-fileserver/tree/master

Right now I am running this on a digital ocean droplet which can be accessed here:

http://165.22.207.94:3030/

might continue messing with it a bit and I wont be surprised if I have a made a few security mistakes with my inexperience :3

### DIFF
--- a/week1/homework/src/server.js
+++ b/week1/homework/src/server.js
@@ -5,16 +5,46 @@ const http = require('http');
 /* `createServer` MUST return an instance of `http.Server` otherwise the tests
  * will fail.
  */
+
+function responseHandler(response, status, state) {
+  response.writeHead(status, { 'Content-Type': 'application/json' });
+  if (status === 200) {
+    response.write(JSON.stringify({ state }));
+  }
+ else {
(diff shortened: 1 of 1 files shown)


week1/homework/src/server.js
```diff
@@ -5,16 +5,46 @@ const http = require('http');
 /* `createServer` MUST return an instance of `http.Server` otherwise the tests
  * will fail.
  */
+
+function responseHandler(response, status, state) {
```
gajduk: I think `responseHandler` is not the best possible name.  Can you come up with a better name?

week1/homework/src/server.js
```diff
@@ -5,16 +5,46 @@ const http = require('http');
...
+  response.writeHead(status, { 'Content-Type': 'application/json' });
```
gajduk: correct status and content type 👍

Tjebbee:
@Laurelinduilwen can you implement @gajduk 's feedback?

PR #438, opened 2019-08-24
Node week 2

elicarcar:
*** **JUST SOME NOTES** ***
- Instead of process, I used yargs, but I can change it and use commander instead.
- By default, help command works and shows the tools as expected but I couldn't execute the help options with 'help' or '--help' command

### DIFF
--- a/week2/homework/package.json
+++ b/week2/homework/package.json
@@ -13,5 +13,9 @@
     "eslint-plugin-node": "^6.0.0",
     "eslint-plugin-promise": "^3.5.0",
     "eslint-plugin-standard": "^3.0.1"
+  },
+  "dependencies": {
+    "commander": "^3.0.0",
+    "yargs": "^14.0.0"
   }
 }
--- a/week2/homework/src/commands.js
+++ b/week2/homework/src/commands.js
@@ -1,13 +1,12 @@
 /* eslint-disable indent */
 const fs = require('fs');
 const chalk = require('chalk');
-const log = console.log;
 let todoList = [];
 
 const list = () => {
   todoList = JSON.parse(fs.readFileSync('todoList.json'));
   if (todoList.length === 0) {
-    console.log(`There is nothing to show on ${todoList}, please add some tasks`);
+    console.log(`There is nothing to show on your to-do list, please add some tasks`);
   }
  else {
     console.log(todoList);
@@ -53,15 +52,15 @@ const reset = () => {
 };
 
 const help = () => {
-  log(chalk.magenta('Welcome to this small cli version 0.0.1'));
-  log(chalk.bgCyan.bold('Commands:'));
-  log(chalk.yellow('help') + ' => Executes all the available commands');
-  log(chalk.yellow('list') + " => show currents to-do's");
-  log(chalk.yellow('add --todo="(your task)"') + '  => adds a task to your list');
-  log(
-    chalk.yellow('remove --todo= --task number--') + " => removes the task that you're targeting"
+  console.log(chalk.magenta('Welcome to this small cli version 0.0.1'));
+  console.log(chalk.bgCyan.bold('Commands:'));
+  console.log(chalk.yellow('help') + ' => Executes all the available commands');
+  console.log(chalk.yellow('list') + " => show currents to-do's");
+  console.log(chalk.yellow('add --todo="<your task>"') + '  => adds a task to your list');
+  console.log(
+    chalk.yellow('remove --todo= <task number>') + " => removes the task that you're targeting"
   );
-  log(chalk.yellow('reset') + ' => resets the todolist');
+  console.log(chalk.yellow('reset') + ' => resets the todolist');
 };
 module.exports = {
   add,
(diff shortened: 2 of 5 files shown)


week2/homework/src/index.js
```diff
@@ -1,3 +1,37 @@
 'use strict';
 
 // TODO: Write the homework code in this file
+const yargs = require('yargs');
+const commands = require('./commands');
+
+let command = yargs.argv._[0];
+let todo = yargs.argv.todo;
```
Meno-101: your solution with `help` option is just to make the following and that will disable it's default value

**Option#1**

you may need to have a new variable to use it's returned value in which it turns off the help command
```javascript
const y = yargs.help(false);

let command = y.argv._[0];
let todo = y.argv.todo;
```

**Option#2**

or you may include it at line 4 as:
```javascript
const yargs = require('yargs').help(false);
```

Since you are not changing their direct value: `command` & `todo` should be a `const` too.

week2/homework/src/index.js
```diff
@@ -1,3 +1,37 @@
...
+      commands.add(todo);
+    }
+    break;
+  case 'remove':
+    if (todo) {
+      commands.remove(todo);
+    }
+    break;
+  case 'list':
+    if (todo) {
```
Meno-101: why you have a `todo` condition if you just want to list everything ?

week2/homework/src/index.js
```diff
@@ -1,3 +1,37 @@
...
+      commands.list();
+    }
+    break;
+  case 'reset':
+    if (todo) {
```
Meno-101: same here!! 🙃

week2/homework/src/commands.js
```diff
@@ -0,0 +1,65 @@
...
+};
+
+const help = () => {
+  log(chalk.magenta('Welcome to this small cli version 0.0.1'));
+  log(chalk.bgCyan.bold('Commands:'));
+  log(chalk.yellow('help') + ' => Executes all the available commands');
+  log(chalk.yellow('list') + " => show currents to-do's");
+  log(chalk.yellow('add --todo="(your task)"') + '  => adds a task to your list');
+  log(
+    chalk.yellow('remove --todo= --task number--') + " => removes the task that you're targeting"
```
Meno-101: people used to read a specific style of documentation,
just to be sure everyone would understand this message as it's, 
you could make it as `<todo number>` instead of `--todo number--` since the `--` is a valid flag in a shell language, others like me would take as a valid thing to add, unless they have a very clear idea that this is a one thing not two separated things

elicarcar:
Hello @Meno-101 !! Thank you for your feedback, I believe i implemented most of them and now my code seems to be working! At least that's what I'm hoping for

week2/homework/package.json
```diff
@@ -13,5 +13,9 @@
     "eslint-plugin-node": "^6.0.0",
     "eslint-plugin-promise": "^3.5.0",
     "eslint-plugin-standard": "^3.0.1"
+  },
+  "dependencies": {
+    "commander": "^3.0.0",
+    "yargs": "^14.0.0"
```
Meno-101: That's really good to see 💐 

but commander isn't in use.

week2/homework/src/commands.js
```diff
@@ -0,0 +1,72 @@
...
+  todoList.splice(index - 1, 1, { str });
+  console.log(`The item in the ${index} place is up-to date`);
+  fs.writeFileSync('todoList.json', JSON.stringify(todoList));
+};
+
+const reset = () => {
+  try {
+    todoList = JSON.parse(fs.readFileSync('todoList.json'));
+    todoList.length = 0;
+    fs.writeFileSync('todoList.json', JSON.stringify(todoList));
```
Meno-101: The only thing this function does is to rewrite the file making it an empty array, that would be great to have but on the other hand when we are removing with no care of what content is, then there is no need for the read on line 45 and line 46.

the solution will be a super easy one after removing lines 45, 46 we could do:
```suggestion
    fs.writeFileSync('todoList.json', JSON.stringify([])); // done :)
```

week2/homework/src/commands.js
```diff
@@ -0,0 +1,72 @@
...
+const add = todo => {
+  try {
+    todoList = JSON.parse(fs.readFileSync('todoList.json'));
+    console.log(`${todo} is added to your todo list`);
+    todoList.push({ todo });
+
+    fs.writeFileSync('todoList.json', JSON.stringify(todoList));
+  }
+ catch (err) {
+    throw Error(err);
```
Meno-101: Throwing an error in a catch may stop your whole program, if it was a server this could be a problem, but sometimes we need, this is fine, but it's not always a good thing to throw an error after catching it, other wise we won't catch:)

Meno-101:
Oh yes and one last thing, `force push` isn't always a good thing, don't do it if you are not sure of what are you doing And never do it in your job unless you are very comfortable with git. 💔

Tjebbee:
@elicarcar , can you have a look at the feedback?

PR #439, opened 2019-08-25
submission Homework week 2

Mehmet-Ergorgec:


week2/homework/src/todo.js
```diff
@@ -0,0 +1,76 @@
+const fs = require('fs');
+
+const loadJSON = () => {
```
gajduk: nice that you extracted this code into separate functions 👍

week2/homework/src/todo.js
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const saveJSON = todos => {
+  const dataJSON = JSON.stringify(todos);
+  fs.writeFileSync('todos.json', dataJSON);
+};
+
+const createOrderValue = array => {
+  array.forEach((elem, index) => {
+    return (elem.order = index + 1);
```
gajduk: where is this return value used?

week2/homework/src/todo.js
```diff
@@ -0,0 +1,76 @@
...
+    return [];
+  }
+};
+
+const saveJSON = todos => {
+  const dataJSON = JSON.stringify(todos);
+  fs.writeFileSync('todos.json', dataJSON);
+};
+
+const createOrderValue = array => {
```
gajduk: This is correct but it feels like an overkill.
If I wanted the element with order 5 from a javascript array I would use `array[5-1]`
The only possible use for order is when printing the list to the user

week2/homework/src/todo.js
```diff
@@ -0,0 +1,76 @@
...
+  createOrderValue(todo);
+  console.log('New to-do added');
+
+  saveJSON(todo);
+};
+
+const removeToDo = order => {
+  const todo = loadJSON();
+
+  const keepToDo = todo.filter(elem => elem.order !== order);
```
gajduk: Your code is correct but it might be simpler to write 
`todo.splice(order-1, 1); ` removes one element at position order-1 from the array

week2/homework/src/todo.js
```diff
@@ -0,0 +1,76 @@
...
+const listToDo = () => {
+  const todo = loadJSON();
+
+  if (todo.length === 0) {
+    console.log('Your list is empty now.');
+  } else {
+    console.log('Your To-Do`s');
+    createOrderValue(todo);
+    todo.forEach(elem => {
+      return console.log(`${elem.order}.${elem.title}`);
```
gajduk: Using order here makes sense and results in a more readable output 👍

week2/homework/src/todo.js
Mehmet-Ergorgec: I think no need to use return value. Thank you

week2/homework/src/todo.js
Mehmet-Ergorgec: cool Thank you!

PR #440, opened 2019-08-26
Adham nodejs homework week2

Adham-Elias-Botours:
This is my homework for week 2 , node.JS
I did not prefer to use `commander `library , because I do not think it is necessary.
On the other hand I handled almost all what the user might type.  

Thanks


week2/homework/src/read.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+const fs = require('fs');
+
+exports.reading = function() {
```
Meno-101: I would keep `read` and `write` functionalities in a one file called `utiles` instead of multiple.

what if you have dozens of features would you keep them in a many files, other than what you necessarily have?

Adham-Elias-Botours: One file called `utiles` , this makes things much better. 
And I agree about descriptive naming issue. 
Thank you for your time on reviewing this homework.

PR #441, opened 2019-08-26
Submit week2 homework - Oguzhan

oguzhanturgut:
Homework `week2/homework`

- Read/write JSON file with sync mode
- Commander and Inquirer modules
- TodoList class ( private methods with WeakMap type for encapsulation)
- Separate files for add, list, remove, update and reset operations
`(add.js, config.js, index.js, list.js, remove.js, reset.js, todo.json, todoList.js, update.js)`

Extra Homework `week2/extra-homework`

- Express module with simple logging middleware
- Async/await in express callbacks
- Async file i/o using  static methods in Counters class (thanks to @remarcmij for the tip)
- CRUD operations over `counters.json` file
`(config.js, counters.js, counters.json, index.js)`


week2/homework/src/todoList.js
```diff
@@ -0,0 +1,77 @@
...
+
+    _read.set(this, file => {
+      const data = fs.readFileSync(file, { encoding: config.ENCODING });
+      return JSON.parse(data);
+    });
+
+    _write.set(this, data => {
+      fs.writeFileSync(this.file, JSON.stringify(data, null, 2));
+    });
+    _todos.set(this, _read.get(this)(this.file));
```
Meno-101: Cool trick 🔥

week2/homework/src/todoList.js
```diff
@@ -0,0 +1,77 @@
...
+    this.todos.push({
+      //* Id will be 1 if the todo list is empty, otherwise 1 greater than the max id
+      id: this.todos.length ? Math.max(...this.todos.map(i => i.id)) + 1 : 1,
+      description: task
+    });
+    _write.get(this)(this.todos);
+  }
+
+  remove(id) {
+    if (!this.todos.map(i => i.id).includes(parseInt(id))) {
```
Meno-101: you could do the following making it easier to read and shorter.
```javascript
    const isExists = this.todos.find(i => i.id === parseInt(id))
    if (!isExists) {
      throw new Error('Invalid Index');
    }
```

Also there is a comment is related to this see the next one.

week2/homework/src/todoList.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  update(id, task) {
+    if (!this.todos.map(i => i.id).includes(parseInt(id))) {
+      throw new Error('Invalid Index');
+    }
+    if (!task.length) {
+      throw new Error('Task required');
+    }
+    this.todos.find(item => item.id === +id).description = task;
```
Meno-101: you have used the same methodology as my previous comment here, perhaps if you used it on line `65` you can then use it's variable to reassign the `.description` here:)

it's all about shorter code, and the fastest implementation we could improve,
your methodology of `.map` then `.include` uses two different loops not mentioning the one here, while this uses a one only in any case, and that's would make some more I/O blocking 😉

week2/homework/src/todoList.js
```diff
@@ -0,0 +1,77 @@
...
+
+  remove(id) {
+    if (!this.todos.map(i => i.id).includes(parseInt(id))) {
+      throw new Error('Invalid Index');
+    }
+    const filteredTodos = this.todos
+      .filter(task => task.id !== id)
+      .map(task => task.description)
+      .reduce((tasks, description, index) => {
+        tasks.push({ id: index + 1, description });
```
Meno-101: usually `id`s doesn't change, once it's on a one, it's just meant to be for that one.

imagine the following scenario when my browser knows it's id as `5` and the value of `301` for some reason and it's been increased or decreased for situation and use case, why should it use the next one if the previous client that uses this website having an event handler that fires on disconnect to remove it's counter, so I may start from `0` or I may not have a counter which in some case could be important

week2/homework/src/add.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+const program = require('commander');
+const todoList = require('./todoList');
+
+program.parse(process.argv);
+
+try {
+  const task = program.args.join(' ');
```
Meno-101: Cool stuff 🎉 

Screen Shot 2019-08-29 at 14 28 36

week2/extra-homework/index.js
```diff
@@ -0,0 +1,85 @@
...
+  } catch (error) {
+    console.log(error.message);
+  }
+});
+
+app.post('/counters', async (req, res) => {
+  try {
+    const counters = await Counters.read(COUNTERS_FILE, config.ENCODING);
+    const id = Number(Math.max(...counters.map(counter => counter.id))) + 1;
+    newCounter = { id, value: 0 };
```
Meno-101: this is a global variable!

week2/extra-homework/index.js
```diff
@@ -0,0 +1,85 @@
+'use strict';
+
+const fs = require('fs');
+const express = require('express');
+const Counters = require('./counters');
```
Meno-101: for this part, the require method assumes it's argument to be a `js`, `json` or a `node` extension, and this line specifically takes the `js` extension because it's the first on the list, but to not be confused, and it's a personal style, what about renaming `counter.json` to be another name?

week2/extra-homework/counters.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const fs = require('fs');
+
+class Counters {
+  constructor(file) {
+    this.file = file;
+  }
+
+  static read(file, encoding) {
```
Meno-101: The encoding is mostly `utf8` in your case since it's counters `json` file on any write or in any case you have.

week2/homework/src/todoList.js
oguzhanturgut: Thank you very much

week2/homework/src/todoList.js
oguzhanturgut: I have tried to reorder the ids that is because `list` also shows the `id`s as indexes. You are right, the `id`s must be unique and should not be mutated also for data integrity and consistency. But in this homework i used the `id`s just for publishing the order of tasks in a sequential look.

I quit reordering the `id`s which also resulted a simpler code  😁

week2/homework/src/todoList.js
oguzhanturgut: I have missed that point and refactored with a new todo variable which can be used for existance checking. Thanks for the tip 🙏

```
update(id, task) {
    const todo = this.todos.find(item => item.id === parseInt(id));
    if (!todo) throw new Error('Invalid Index');
    if (!task.length) throw new Error('Task required');
    todo.description = task;
    _write.get(this)(this.todos);
  }
```

week2/homework/src/todoList.js
oguzhanturgut: Refactored as the `update` method below

```
remove(id) {
    const todo = this.todos.find(item => item.id === parseInt(id));
    if (!todo) throw new Error('Invalid Index');
    const filteredTodos = this.todos.filter(task => task.id !== id);
    _write.get(this)(filteredTodos);
  }
```

week2/homework/src/add.js
oguzhanturgut: Thanks  😃

week2/extra-homework/index.js
oguzhanturgut: 🙈 I forgot to put my precious `const`

week2/extra-homework/index.js
oguzhanturgut: Yeah there is definitely a counter naming confusion. Everywhere is full of counters 😁
Changed the json file name to counter

week2/extra-homework/counters.js
oguzhanturgut: That is right, Hard coded the `fs.readFile` to `utf8`

oguzhanturgut:
> Hello @oguzhanturgut
> 
> Amazing, thanks for you homework, you did the extra one too, I guess you have a programming background?
> 
> and thanks @remarcmij for the help ❤️💐

Hello @Meno-101 
Thanks for your time and valuable feedback. I have some background on IT but not too much in programming. I think i am at the very beginning of a long programming journey.

Meno-101:
Thanks for the update, that explains a lot, it looks a quit interesting journey.

And just something I haven't mentioned, don't forgot to handle your errors in the future. 🙂

oguzhanturgut:
Thanks for your kind comments again. For the error handling,  i will always keep your tip i mind 🙏

PR #442, opened 2019-08-26
Mustafa homework2

mayqf:


week2/homework/src/index.js
```diff
@@ -1,3 +1,85 @@
 'use strict';
+const fs = require('fs');
+const DEFAULT_ENCODING = 'utf8';
+const FILE_PATH = 'c:/nodeJS/week2/homework/src/todo.txt';
```
Meno-101: this will not work on all devices, but windows only and only if the user have the project in that exact location on there devices!

instead you do the following:
```suggestion
const FILE_PATH = __dirname + '/todo.txt';
```
this should work just fine in any platform.

Screen Shot 2019-08-29 at 15 25 35

week1/homework/src/server.js
```diff
@@ -1,15 +1,45 @@
 'use strict';
```
Meno-101: Hmm, this is the first home work I guess you accidentally made your new branch out of the week1 branch making them follow each other history.

for avoiding that in the future, you should make checkout to master first, make a new branch, then you ca check out to it.
```bash
git checkout master
git branch <new-branch> #name it as you wish
git checkout <new-branch>
```
Or you could possibly as a shortcut do this by one command
```bash
git checkout master
# `-b` option will make a new branch and it will switch to it.
git checkout -b <new-branch> #name it as you wish
```
Or even shorter, with online you could do it as:
```bash
# `-b` option will make a new branch and it will switch to it.
# new-branch it could be `week2` and source-branch could be `master` in your case because we need to checkout from the master branch.
git checkout -b <new-branch> <source-branch>
```

Meno-101:
Hello @mayqf,
thanks for your homework, it's short one.

you didn't do the extra part and any bonus.

But, Good job.

week2/homework/src/index.js
mayqf: ok. So __dirname defines file path globally. So it works all other devices. Not only speciliazed for my computer?

week1/homework/src/server.js
mayqf: There is not explanation on this week's assignment, so i did not care it. I will consider it in my next homework.Thanks

week2/homework/src/index.js
Meno-101: yup!

PR #443, opened 2019-08-27
w2-homework-Shyar

ShyarK:


Meno-101:
Nope it's my bad, sorry for bothering.

week1/homework/src/fileContent.js
```diff
@@ -0,0 +1,13 @@
+function State(num) {
```
Meno-101: Hmm, this is the first home work I guess you accidentally made your new branch out of the week1 branch making them follow each other history.

for avoiding that in the future, you should make checkout to master first, make a new branch, then you ca check out to it.
```bash
git checkout master
git branch <new-branch> #name it as you wish
git checkout <new-branch>
```
Or you could possibly as a shortcut do this by one command
```bash
git checkout master
# `-b` option will make a new branch and it will switch to it.
git checkout -b <new-branch> #name it as you wish
```
Or even shorter, with online you could do it as:
```bash
# `-b` option will make a new branch and it will switch to it.
# new-branch it could be `week2` and source-branch could be `master` in your case because we need to checkout from the master branch.
git checkout -b <new-branch> <source-branch>
```

week2/homework/src/commands.js
```diff
@@ -0,0 +1,10 @@
+const fs = require('fs');
+const colors = require('colors');
```
Meno-101: First Error appears when I launched the application was `module not found`

when you are adding a new module using `npm` make sure to use `--save` option to save it as a dependency that will make any one using your program prepare it with a single command `npm install` not warring about what modules he have to include, what if you have around 10 to 20 module and each should be in it's specific version to run as it's right way?

week2/homework/src/index.js
```diff
@@ -1,3 +1,52 @@
 'use strict';
 
 // TODO: Write the homework code in this file
+const program = require('commander');
```
Meno-101: same thing here!

week2/homework/src/index.js
```diff
@@ -1,3 +1,52 @@
...
+    .action(() => {
+      showList.showToDosList();
+    });
+
+  program
+    .command('remove <index>')
+    .alias('rm')
+    .description('Removes a to-do item')
+    .action(index => {
+      remove.removeList(index);
```
Meno-101: Exactly it removes `Removes a to-do item` your method called `removeList`.

week2/homework/src/index.js
```diff
@@ -1,3 +1,52 @@
...
+    .action(() => {
+      commands.showCommandsList();
+    });
+
+  program
+    .command('add <toDo>')
+    .alias('a')
+    .description('Adds a to-do item')
+    .action(toDo => {
+      addToDoList.generateToDosList(toDo);
```
Meno-101: Here also it `Adds a to-do item` your method called `addToDoList.generateToDosList`.

week2/homework/src/removeList.js
```diff
@@ -0,0 +1,19 @@
+const fs = require('fs');
+const colors = require('colors');
+function removeList(index) {
+  try {
+    if (index <= 0 || index > toDosArray.length) {
```
Meno-101: when I launched the command:

```bash
node index.js remove 1
```

I got this error:

```
ReferenceError: Cannot access 'toDosArray' before initialization
```

week2/homework/src/removeList.js
```diff
@@ -0,0 +1,19 @@
...
+      console.log(`You've got ${toDosArray.length} list/s`);
```
Meno-101: this will show the message but it won't stop, but it will try to remove something isn't exist or it's not valid, so it may give an error.

ShyarK: I didn't get this point! Can you clarify it again? Maybe you meant to get if statement out of try?

week1/homework/src/fileContent.js
ShyarK: I think I'll go to do it again with a new pull request!

Meno-101: No need just be aware of it the next one:)

week2/homework/src/removeList.js
Meno-101: I mean you have an invalid index, and you are making that clear to the user by telling him how many items he have.

and then you are reading the file on line 8 and taking of the item on line 9 even it is not valid item and it will be an `undefined` value, and you completing the process till writing `undefined` to the array.

ShyarK: I think I fixed the problem! 
try {
    const toDosArray = JSON.parse(fs.readFileSync('toDoList.json'));
    const item = toDosArray[index - 1];
    const lists = toDosArray.filter(list => list !== item);
    fs.writeFileSync('./toDoList.json', JSON.stringify(lists));
    if (index <= 0 || index > toDosArray.length) {
      console.log(`You've got ${toDosArray.length} list/s`);
    }
  } catch (error) {
    console.log(colors.yellow(`${error}`));
  }

week2/homework/src/commands.js
ShyarK: Just to clarify, inside src folder I have, addToDos, commands.js, index.js, removeList.js, resetList.js and showList.js files. Now I need inside src folder run `npm init` right?

Meno-101: you have a related `package.json` exist already with the main key it have, so you may just have to install the packages there.

once you have that done, who ever is using the program will be notified to install any packages with `npm install` command only once he have your project from reading the `package.json` file.

week2/homework/src/removeList.js
Meno-101: that will solve the error you have but it will stay writing unnecessary `undefined` to the file you have and it may cause an error since it should be writing something to the file.

avoiding any unnecessary stuff you may do the following:
```javascript
try {
  const toDosArray = JSON.parse(fs.readFileSync('toDoList.json'));
  if (index <= 0 || index > toDosArray.length) {
    console.log(You've got ${toDosArray.length} list/s);
  } else {
    // this will prevent any redundant actions
    const item = toDosArray[index - 1];
    const lists = toDosArray.filter(list => list !== item);
    fs.writeFileSync('./toDoList.json', JSON.stringify(lists));
  }
} catch (error) {
  console.log(colors.yellow(${error}));
}
```

week2/homework/src/index.js
```diff
@@ -1,3 +1,53 @@
...
+const program = require('commander');
+const commands = require('./commands');
+const addToDo = require('./addToDo');
+const showList = require('./showList');
+const remove = require('./removeToDo');
+const reset = require('./resetList');
+if (process.argv.length === 2) {
+  commands.showCommandsList();
+}
+ else {
```
Meno-101: this is somehow a long `else` close, perhaps we could simplify it by looking to it's actions:
 - we want a help on `help` command
 - we want a help on an empty submit

from the first we could:
```javascript
program.addImplicitHelpCommand = () => {};
```
And the second: 
```javascript
if (process.argv.length === 2) {
  process.argv.push('help')
}
```

references: [first], the [second].

[first]: https://github.com/tj/commander.js/issues/683#issuecomment-339602122
[second]: https://github.com/tj/commander.js/issues/303#issuecomment-477441964

week2/homework/src/addToDo.js
```diff
@@ -0,0 +1,20 @@
+const fs = require('fs');
+let toDoList = [];
+function generateToDo(toDo) {
+  try {
+    toDoList = JSON.parse(fs.readFileSync('toDoList.json'));
+  }
+ catch (error) {
+    console.log(`${error}`);
```
Meno-101: perhaps you need to move these line from 10 to 15 to the try close. :)
when I have added a new item I got: `Error: ENOENT: no such file or directory, open 'toDoList.json'`

PR #447, opened 2019-08-28
Week2 Homework- Salih S

salih18:
-  'yargs' package to build the cli app and to parse the process.argv

-  'readline' package in order to respond to the user interactively and a user friendly.

- 'chalk' package for the styling

- 'path' package to address the file

- 'fs promises' with async await usage which has been added Node Js since v10. 

	
### **App Features: v1.0**

  Usage: **index.js command**

  **index.js list**      List to-dos, if there is no to-do then show the information about it.
  **index.js add**       Add a new to-do with the use of recursive function that checks todo has 5 letters long.
  **index.js remove**    Remove a to-do by asking the order of the todo
  **index.js complete**  Complete a to-do then checks it as a completed. Red label shows uncompleted, green shows completed todo.
  **index.js reset**     Remove all to-dos by asking to continue yes/no confirmation with recursive funct
  **index.js read**      Read a specific to-do.
  **index.js update**    Update a specific to-do.
  **index.js export**    Export to-dos as a .txt file by asking the user the name of the file, and then shows where it has been exported.
  **index.js search**    Search for a term in to-dos based on including keywords by case insensitive.
  **index.js help**      Show the usage of the app 

Note: I had to update the Eslint conf file because it causes indentation errors in my project. I copied the first week conf file to fix the problem.

### DIFF
--- a/week2/homework/src/exportToDos.js
+++ b/week2/homework/src/exportToDos.js
@@ -9,10 +9,7 @@ const exportToDos = async filename => {
   try {
     const todos = await loadToDos();
     const output = todos
-      .map(todo => {
-        const line = [todo.index + '. ' + todo.body + ' ' + todo.complete];
-        return line.join(', ');
-      })
+      .map(todo => [todo.index + '. ' + todo.body + ' ' + todo.complete].join(', '))
       .join('\n');
     fs.writeFile(file + '.txt', 'Your to-dos:\n' + output).then(() => {
       logScreen(`Your todos exported into ${file}.txt file`, 'green');
--- a/week2/homework/src/exportToDos.js
+++ b/week2/homework/src/exportToDos.js
@@ -9,12 +9,13 @@ const exportToDos = async filename => {
   try {
     const todos = await loadToDos();
     const output = todos
-      .map(todo => [todo.index + '. ' + todo.body + ' ' + todo.complete].join(', '))
+      .map(todo => [todo.index + '. ' + todo.body + ' ' + todo.complete])
       .join('\n');
     fs.writeFile(file + '.txt', 'Your to-dos:\n' + output).then(() => {
       logScreen(`Your todos exported into ${file}.txt file`, 'green');
     });
-  } catch (err) {
+  }
+ catch (err) {
     logScreen(`Could not been exported! Please, try again...`, 'red');
   }
 };
--- a/week2/homework/src/exportToDos.js
+++ b/week2/homework/src/exportToDos.js
@@ -1,0 +1,25 @@
+'use strict';
+
+const { loadToDos, logScreen } = require('./Util');
+const path = require('path');
+const fs = require('fs').promises;
+
+const exportToDos = async filename => {
+  const file = path.join(__dirname, filename);
+  try {
+    const todos = await loadToDos();
+    const output = todos
+      .map(todo => {
+        const line = [todo.index + '. ' + todo.body + ' ' + todo.complete];
+        return line.join(', ');
+      })
+      .join('\n');
+    fs.writeFile(file + '.txt', 'Your to-dos:\n' + output).then(() => {
+      logScreen(`Your todos exported into ${file}.txt file`, 'green');
+    });
+  } catch (err) {
+    logScreen(`Could not been exported! Please, try again...`, 'red');
+  }
+};
+
+module.exports = exportToDos;
--- a/week2/homework/src/exportToDos.js
+++ b/week2/homework/src/exportToDos.js
@@ -15,7 +15,7 @@ const exportToDos = async filename => {
       logScreen(`Your todos exported into ${file}.txt file`, 'green');
     });
   }
- catch (err) {
+  catch (err) {
     logScreen(`Could not been exported! Please, try again...`, 'red');
   }
 };
--- a/week2/homework/src/index.js
+++ b/week2/homework/src/index.js
@@ -1,8 +1,20 @@
 'use strict';
 
+const {
+  prompt,
+  addPrompt,
+  removePrompt,
+  completePrompt,
+  resetPrompt,
+  readPrompt,
+  updatePrompt,
+  updateQuestion,
+  exportPrompt,
+  searchPrompt
+} = require('./prompt');
+
 const chalk = require('chalk');
 const yargs = require('yargs');
-const rl = require('readline').createInterface(process.stdin, process.stdout);
 const listToDos = require('./listToDos');
 const addToDo = require('./addToDo');
 const removeToDo = require('./removeToDo');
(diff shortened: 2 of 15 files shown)


week2/homework/src/index.js
```diff
@@ -1,3 +1,126 @@
...
+      rl.question(chalk.yellow(`Please, enter to-do order to complete it:  `), answer => {
+        completeToDo(parseInt(answer));
+        rl.close();
+      });
+    },
+  })
+  .command({
+    command: 'reset',
+    describe: 'Remove all to-dos',
+    handler: function() {
```
Meno-101: for the handlers too, it's one of the things we should be doing all the time, sometimes making it easier to debug and cleaner for the eye to read, we should move all these handlers to a isolated file that will take care of all question and lins we need to deal with.

then exporting them will give us the space here to see what kinf of commands we have with out running the help command from the command line, think about this: what if the `help` command is broken for some reason how can i know how your program works if I'm not a javascript person?

week2/homework/src/index.js
```diff
@@ -1,3 +1,126 @@
 'use strict';
 
-// TODO: Write the homework code in this file
+const chalk = require('chalk');
+const yargs = require('yargs');
+const rl = require('readline').createInterface(process.stdin, process.stdout);
```
Meno-101: It's not always needed the use case of `readline`, if I ran the command `node index.js` with out any other command it will stuck.

Also line 22 you are just closing it.
so you may create interface on every handler or you could create your own function using `Promise` and async await on every handler, it will be a good exercise :)

week2/homework/src/exportToDos.js
```diff
@@ -0,0 +1,22 @@
...
+const { loadToDos, logScreen } = require('./Util');
+const path = require('path');
+const fs = require('fs').promises;
+
+const exportToDos = async filename => {
+  const file = path.join(__dirname, filename);
+  try {
+    const todos = await loadToDos();
+    const output = todos
+      .map(todo => [todo.index + '. ' + todo.body + ' ' + todo.complete].join(', '))
```
Meno-101: Some of these are not needed, I think you may don't even use them and they will work just fine:
```suggestion
      .map(todo => todo.index + '. ' + todo.body + ' ' + todo.complete)
```

week2/homework/src/resetToDos.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+const { loadToDos, saveToDo, logScreen } = require('./Util');
+
+const resetToDos = async () => {
+  const todos = await loadToDos();
```
Meno-101: no need to read any todos here, you just have to set the todos at write to be an empty array, same result and simpler way.

week2/homework/src/updateToDo.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const { loadToDos, saveToDo, logScreen } = require('./Util');
+
+const updateToDo = async (order, body) => {
+  const todos = await loadToDos();
+  const todo = todos.find(todo => todo.index === order);
+  if (todo) {
+    todos[order - 1] = { index: order, body, complete: '➖' };
```
Meno-101: you could instead make it as what you have done with [completeTodo]:
```javascript 
todos[order - 1].body = 'stuff'
```

other than legibility, what if this todo is done, you are resetting it to be not done.

[completeTodo]: https://github.com/HackYourFuture/Node.js/pull/447/files#diff-c03001336cfcb3bf1db5d7fb92e1193aR8

Meno-101:
Oh yes and please have a look to the comments and apply the changes :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,126 @@
...
+    describe: 'Search for a term in to-dos',
+    handler: function() {
+      rl.question(chalk.yellow(`Please, enter a keyword :  `), answer => {
+        searchToDos(answer.toLowerCase());
+        rl.close();
+      });
+    },
+  })
+  .usage('Usage: $0 <command>')
+  .epilog(chalk.yellow('Copyright 2019 >> salih18.github.io'))
```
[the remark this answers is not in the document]
salih18: Thank you...

week2/homework/src/exportToDos.js
salih18: I think,I forgot that while trying to accomplish.. Thanks

week2/homework/src/resetToDos.js
salih18: Yes, absolutely you are right. Thanks.

week2/homework/src/updateToDo.js
salih18: I have noticed it now, that is why reviewing the code by someone else important. :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,126 @@
...
+      });
+    },
+  })
+  .command({
+    command: 'reset',
+    describe: 'Remove all to-dos',
+    handler: function() {
+      rl.question(
+        `${chalk.red.inverse('This will remove all to-dos!!!')}\n
+      Type ${chalk.green('yes')} to continue or ${chalk.red('no')} to cancel\n`,
```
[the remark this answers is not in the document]
salih18: I move it to separate file.

week2/homework/src/index.js
salih18: I move it to separate file. Besides I add some methods to yargs, whether or not if the user typed anyother commands which are not included, it says to the user, that here is the help and usage of this app... :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,114 @@
...
+    builder: {
+      complete: {
+        describe: 'To-do complete',
+        type: 'string',
+        default: '➖'
+      }
+    },
+    handler: async function(argv) {
+      const answer = await prompt(addPrompt);
+      answer.length >= 5 ? addToDo(answer, argv.complete) : this.handler(argv);
```
Meno-101: This complete feature isn't working properly unless the user have ✔emoji, which isn't always possible on all platforms.

I would change line 43 to be an empty string, and once the user inserting something as `--complete=true` it will be affected to be the emoji you want ✔. :)

week2/homework/src/index.js
```diff
@@ -1,3 +1,114 @@
...
+      readToDo(parseInt(answer));
+    }
+  })
+  .command({
+    command: 'update',
+    describe: 'Update a specific to-do',
+    handler: async function() {
+      const order = await prompt(updatePrompt);
+      const body = await prompt(updateQuestion);
+      updateToDo(parseInt(order), body);
```
Meno-101: This update function sets the complete `false` if it's `true`

week2/homework/src/prompt.js
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable max-len */
+'use strict';
+
+const chalk = require('chalk');
+const rl = require('readline');
+
+const prompt = question => {
```
Meno-101: I liked this function. ❤️

PR #449, opened 2019-08-29
Submit Week2 extra homework -- Salih S.

salih18:
1. **fs promises** package to read a file asyncronously
2. **path package** to adress a file
3. **body parser** middleware to access request body

I thought that new counter object should come via post request in the body so that I can parse it by using body parser. I could have implemented a function that creates a counter, however I prefer using Postman to send a new counter in the body to simulate post request. e.g {"id":60 }

Besides I want to say that I havent implemented the error handling situation yet which includes error responses about http response codes, error information for the developers etc..


week2/homework/src/counter.js
```diff
@@ -0,0 +1,74 @@
...
+
+const saveCounters = async counters => {
+  try {
+    return await fs.writeFile(FILE, JSON.stringify(counters));
+  } catch (err) {
+    console.log(err);
+  }
+};
+// This function might be not needed
+const listCounters = async () => {
```
Meno-101: Yup!

week2/homework/src/counter.js
```diff
@@ -0,0 +1,74 @@
...
+app
+  .get('/counters', async (req, res) => {
+    res.send(await listCounters());
+  })
+  .get('/counters/:id', async (req, res) => {
+    res.send(await readCounter(parseInt(req.params.id)));
+  })
+  // The post request coming from postman in the body as a JSON. e.g {"id":90 }
+  .post('/counters', async (req, res) => {
+    res.send(await addCounter(parseInt(req.body.id)));
```
Meno-101: a user should not have a control over the `id`, I would prefer to go for using this `post` method with no body so it sets the counter and returns it as a response and from there we can use it's `id`

week2/homework/src/counter.js
```diff
@@ -0,0 +1,74 @@
...
+};
+
+const readCounter = async id => {
+  const counters = await loadCounters();
+  return counters.find(counter => counter.id === id);
+};
+
+const updateCounter = async id => {
+  const counters = await loadCounters();
+  const counter = counters.find(counter => counter.id === id);
```
Meno-101: you can use `readCounter` function here instead of these two lines.

week2/homework/src/counter.js
```diff
@@ -0,0 +1,101 @@
...
+  } catch (e) {
+    return [];
+  }
+};
+
+const listCounters = async () => {
+  try {
+    return await loadCounters();
+  } catch (e) {
+    throw new Error('Please, refresh the page!');
```
Meno-101: you are throwing an error in your catch statement,  this may cause a server failure and it will make it stop in nodejs if for some reason you forgot to catch your errors.

week2/homework/src/counter.js
```diff
@@ -0,0 +1,101 @@
...
+      res.send(await listCounters());
+    } catch (e) {
+      res.status(400).send(e.message);
+    }
+  })
+  .get('/counters/:id', async (req, res) => {
+    try {
+      res.send(await readCounter(parseInt(new Number(req.params.id))));
+    } catch (e) {
+      res.status(400).send(e.message);
```
Meno-101: this could be 404 not found, and line 92 too since both has the message of not found as a response.

week2/homework/src/counter.js
```diff
@@ -0,0 +1,101 @@
...
+  saveCounters(counters);
+  return counter;
+};
+
+app
+  .get('/counters', async (req, res) => {
+    try {
+      res.send(await listCounters());
+    } catch (e) {
+      res.status(400).send(e.message);
```
Meno-101: you may send here a 500 error since it could be a server file reading failure may be?

PR #450, opened 2019-08-29
week2/Adil

musafir928:
… >6.reset.js >7.list.js


week2/homework/src/add.js
```diff
@@ -0,0 +1,20 @@
+const fs = require('fs');
+
+module.exports = (newToDo) => {
+  const content = fs.readFileSync('./todo-list.json');
+  const data = JSON.parse(content);
+  let id;
```
Meno-101: perhaps this could take a ternary operator as :
```javascript
const id = data.ToDos == 0 ? 1 : data.ToDos.length + 1;
```

PR #455, opened 2019-09-03
Node.js Week 3 Homework

gulyener:


week3/homework/src/index.js
```diff
@@ -1,3 +1,137 @@
...
+
+  // Add a todo and todo.description in req.body to create todos
+  async function createTodo(req, res) {
+    const todo = parseAndValidate(req);
+    const todos = await readTodosFromFile();
+    todo.id = uuid.v4();
+    todo.isDone = false;
+    todos.push(todo);
+    await saveTodos(todos);
+    markAsNotDone(req, res);
```
gajduk: Can you explain why are you calling markAsNotDone here?

week3/homework/src/index.js
```diff
@@ -1,3 +1,137 @@
...
+    const todos = await readTodosFromFile();
+    const todo = todos.find(todo1 => todo1.id === req.params.id);
+    const indexToDelete = todos.indexOf(todo);
+    todos.splice(indexToDelete, 1);
+    await saveTodos(todos);
+    res.status = 200;
+    res.end('Task is deleted');
+  }
+
+  // Homework starts here
```
gajduk: thank you 👍

week3/homework/src/index.js
gulyener: Oops I forgot to delete that thanks for telling me:) I was trying to make todos 'not done' when they were created. At first I tried to use this function but then I said 'What am I doing? There is a much simpler way.'  And so I forgot to delete that line.

week3/homework/src/index.js
gulyener: Thanks for your review. It was easier this time because I had some code to begin with:)

PR #459, opened 2019-09-05
Todos-homework-w3/Shyar

ShyarK:


week3/homework/src/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "src",
+  "version": "1.0.0",
+  "description": "",
+  "main": "app.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
Meno-101: since you don't have an `index.js` this will be a useful tip for the future, we can write our own command to run it easier:
```suggestion
    "start": "node app.js"
```
now with one command you can run your application as `npm start`
the left side is the script name, the right side is what ever you run in your terminal which is `node app.js`

week3/homework/src/api.js
```diff
@@ -0,0 +1,91 @@
...
+    try {
+      const todos = await util.read();
+      const todo = {
+        description: req.body.todo.description,
+        id: uuid.v4(),
+        done: false,
+      };
+      todos.push(todo);
+      await util.save(todos);
+      res.end('todo was successful created!');
```
Meno-101: responding with a unique type of data will make it easier to interact with in the client!

week3/homework/src/api.js
```diff
@@ -0,0 +1,91 @@
...
+    } catch (err) {
+      res.end(err);
```
Meno-101: This could send some information no need for the user to have any knowledge about
perhaps it's best to have a hard coded message instead as: `{ error: { type: "server error" , description: "Oops! something went wrong." } }`

and don't forgot the status code.

week3/homework/src/api.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  async updateTodo(req, res) {
+    try {
+      let todos = await util.read();
+      const todo = todos.find(todo => todo.id === req.params.id);
```
Meno-101: this line has been repeated in many places it would be nice to have it in it's method as:
```javascript
getTodo(id) {
    return util.todos.findIndex(todo => todo.id === id);
}
```

and where you use this you will pass only the `id` as:
```javascript
const index = this.getTodo(req.params.id)
```

on each method I read I thought the next is going to be a different `.find` method.

week3/homework/src/api.js
```diff
@@ -0,0 +1,91 @@
...
+      res.status = 201;
+      res.end('task successfully deleted!');
+    } catch (err) {
+      res.end(err);
+    }
+  }
+
+  async deleteTodos(req, res) {
+    try {
+      let todos = await util.read();
```
Meno-101: perhaps a `let` here is a correct use, but the others you have can be `const`, it's preferred since it uses less memory.

week3/homework/src/api.js
```diff
@@ -0,0 +1,91 @@
...
+      todos = [];
+      await util.save(todos);
```
Meno-101: this is no need to read the file if you are just resetting it to be `[]` you can write it here!
```suggestion
      await util.save([]);
```

week3/homework/src/api.js
```diff
@@ -0,0 +1,91 @@
...
+      res.end(err);
+    }
+  }
+
+  async checkTask(req, res, done, task) {
+    try {
+      const todos = await util.read();
+      const todo = todos.find(todo => todo.id === req.params.id);
+      if (!todo) return res.status(404).send('The task with the given Id was not Found!');
+      done ? (todo.done = true) : (todo.done = false);
```
Meno-101: this could be as:
```javascript
      todo.done = done ? true : false;
```

But in any case you are sending the `boolean` from it's call so it can be:
```suggestion
     todo.done = done;
```

week3/homework/src/validator.js
```diff
@@ -0,0 +1,21 @@
+const Joi = require('@hapi/joi');
```
Meno-101: Cool stuff 🔥

week3/homework/src/validator.js
```diff
@@ -0,0 +1,21 @@
...
+          .trim()
+          .required(),
+      }),
+    });
+    return Joi.validate(todo, schema);
+  };
+  this.userValidator = function(req, res, next) {
+    const { error } = validate(req.body);
+    if (error) return res.status(400).send(error.details[0].message);
+    else next();
```
Meno-101: very good use of `next`!

PR #461, opened 2019-09-06
Proposal for first homework

gajduk:
Instead of building REST APIs and stuff let students build a real web server that serves html and javascript. This will teach them how the browser and server interact with each other.

What do you think @NoerGitKat ?

### DIFF
--- a/week1/homework/README.md
+++ b/week1/homework/README.md
@@ -14,7 +14,38 @@ By doing this homework you will learn:
 
 Create an HTTP web server using the native Node.js `http` module.
 
-TO BE CONTINUED
+1. The server will be a functioning web server that can serve a very simple web site. When opening the server url in the browser e.g. `http:\\localhost:3000` the server needs to serve the following html:
+
+```html
+<html>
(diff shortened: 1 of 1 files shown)


Adham-Elias-Botours:
As a student I would like to do that , that is more interesting than working just with JSON.
But for that I prefer to restructure the Node JS lessons . And 1 more week makes thing much better. 
we might focus n how to serve HTML pages manually, or we might dig more in Express to understand how to use it to build and serve a compleet API with GUI on the client side. 
How to use other people code "Express" is a great skill, then we can go on with modifying and adding things to other people code, like the Lecture folder in the homework. 
On the other hand ,  I think HYF tries to give us "the students" a kind of break between java-script 3 and React, by simpel things with NodeJS. For me That is not so simpel at all, and I have always a desire for a break. 
Thanks.

gajduk:
Here is the feedback from Mauro: This is great step needed! Thanks Andrew!
I agree that fs doesn’t need to be part of Week one :)

PR #462, opened 2019-09-06
Adham node js homework week3

Adham-Elias-Botours:
Here is my Node Js homework week 3:

- My work is in the homework folder. 

- Start file : index.js

- I did not adjusted anything in lecture folder, but it was good to have a look on it.

- The folders and files structure could be better.

- There is lots of code for handling errors, it sound more than the code that gives the expected results.

- HTTP status codes are not that correct, but I avoided using 204, because it sends empty response to the client. While my API sends almost always response with content  to the client. 

- I planed to go on and work on lecture folder to add 4 function to that API. But I stopped at this point. 

- Sure there might be more mistakes, and I hope that they do not effect approving this homework.

Thanks


week3/homework/src/data/data_dealer.js
```diff
@@ -0,0 +1,23 @@
...
+      if (err) reject(err);
+      else resolve(result);
+    });
+  });
+};
+
+const write = function(directory, data) {
+  return new Promise(function(resolve, reject) {
+    fs.writeFile(directory, data, err => {
+      if (err) reject(err);
```
Meno-101: this either needs an `else` close or a return statement in your `if` condition preventing the `resolve` from being called.

week3/homework/src/utilities.js
```diff
@@ -0,0 +1,141 @@
...
+
+const dataDealer = require('./data/data_dealer');
+const read = dataDealer.read;
+const write = dataDealer.write;
+const shortid = require('shortid');
+const TodoItem = require('./todo_obj_constructor');
+
+const validation = function(req) {
+  const reqBodyLength = Object.keys(req.body).length;
+  const reqToDo = req.body.todo || {};
```
Meno-101: instead of all this we can do the following:

```javascript
const reqBodyLength = Object.keys(req.body).length
let reqDescription = reqToDo.description

if (reqBodyLength) {
  reqDescription = reqDescription.trim()
} else {
  reqDescription = ''
}
return reqDescription // your function and this if statement work the same but shorter version.
```

Some comments on this function may approve my thought:
 - line 19: what if the user sent accidentally another key?
 - line 20: these can't be equal! and always fail. please look at this article about [object equality]
 - lines 21 and 22 are a correct condition in your case.

[object equality]: http://adripofjavascript.com/blog/drips/object-equality-in-javascript.html

week3/homework/src/utilities.js
```diff
@@ -0,0 +1,141 @@
...
+  ) {
+    return (reqDescription = '');
+  }
+  return reqDescription;
+};
+
+const findToDo = function(CurrentList, reqId) {
+  let toDo = CurrentList.find(todo => todo.id === reqId);
+  if (toDo === undefined) {
+    throw new Error(`The To-Do item with ID:${reqId} is not existed`);
```
Meno-101: in some case throwing could result in shutting the nodejs server application down!

week3/homework/src/utilities.js
```diff
@@ -0,0 +1,141 @@
...
+  return toDo;
+};
+
+const createToDo = function(req, res) {
+  const toDoDescription = validation(req);
+  if (toDoDescription === '') {
+    res.status(202).json({ Notification: 'The posted To-Do is not valid' });
+    return;
+  }
+  const todoObject = new TodoItem(shortid.generate(), toDoDescription, false);
```
Meno-101: the `id` could be generated inside the `todo_obj_constructor.js` file since it's only related to id:)

week3/homework/src/utilities.js
```diff
@@ -0,0 +1,141 @@
...
+  read('./data/todolist.json', 'utf8')
+    .then(result => {
+      const CurrentList = JSON.parse(result);
+      CurrentList.push(todoObject);
+      const updatedList = JSON.stringify(CurrentList, null, 2);
+      write('./data/todolist.json', updatedList);
+      res.status(201).json({ Notification: 'new To-DO is added' });
+    })
+    .catch(err => res.status(500).json({ Error: err.message }));
```
Meno-101: the `read` method errors mostly are going to be about a the file we have, which means nothing for the user

I would rather hard code the error to be something as `Oops something went wrong!` but generally speaking your errors here sometimes are about existence of the todo so you could pass the `response` object to send the response to the client from the `findTodo` function instead of throwing.

Or as a solution you could pass the `next` method and let it pass to the error handler route with it's correct status code.

week3/homework/src/utilities.js
```diff
@@ -0,0 +1,141 @@
...
+      const wantedTodo = findToDo(CurrentList, reqId);
+      wantedTodo.description = toDoDescription;
+      write('./data/todolist.json', JSON.stringify(CurrentList, null, 2));
+      res.status(201).json({ Notification: 'The To-Do item is modified' });
+    })
+    .catch(err => res.status(500).json({ Error: err.message }));
+};
+
+const deleteToDo = function(req, res) {
+  {
```
Meno-101: 🤔 extras
```javascript
{
  // code here why?
}
```

week3/homework/src/app.js
```diff
@@ -0,0 +1,56 @@
...
+app.post('/todos/:id/done', (req, res) => utilities.modifyToDoStatus(req, res, true));
+
+// ( 8 )
+app.delete('/todos/:id/done', (req, res) => utilities.modifyToDoStatus(req, res, false));
+
+// errors handling: JSON parse & stringify, and invalid endpoints
+app.use((req, res, next) => {
+  const error = new Error('Not found');
+  error.status = 404;
+  next(error);
```
Meno-101: Cool Stuff 🔥

week3/homework/src/app.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
+
+const express = require('express');
+const app = express();
+const utilities = require('./utilities');
```
Meno-101: I would change the `utilities` file name to something like `api`, since it's very related to the api you have.

utilities is more generic name, that contains functions and methods about many things and can be used in many places! such as what you have in `data_dealer.js` file :)
which leads to `data_dealer.js` could be the `utilities.js`?

week3/homework/src/app.js
```diff
@@ -0,0 +1,56 @@
...
+
+/* before dealing with the endpoints:  parse the request body */
+app.use(express.json());
+
+// ( 0 )
```
Meno-101: if these are clarifying for you it's fine during your development, but once you done removing them will give more space for others and less lines to read, image a server with 100 route?

week3/homework/src/utilities.js
```diff
@@ -0,0 +1,141 @@
...
+};
+
+const showToDos = function(req, res) {
+  read('./data/todolist.json', 'utf8')
+    .then(result => {
+      if (result === '[]') {
+        res.status(202).json({ Notification: 'The ToDo list is empty!' });
+        return;
+      }
+      res.status(200).send(result);
```
Meno-101: what about a content type?

week3/homework/src/data/data_dealer.js
Adham-Elias-Botours: I think that I tested it while I have been writing the code. If there is an error then  the code execution will stop after throwing aan error. And for being in the safe side `Else` have to be used.
But Yes, the syntax  of` if statement`  was  confusing for me at that point.

week3/homework/src/utilities.js
Adham-Elias-Botours: - about your point line 19: if the user sends 2 ToDo object :
`{"todo:{"description":"task 1"} ,{"todo:{"description":"task 2"}`}`
here something wired going to happen: first that is not valid JSON, duplicated keys. Second the both ToDos wil be added correctly to the data.

- About your point line 20: in line 11 and 12 I assign the value `{}` to `reqDescription` when something is not defined. That is something wrong I made. I forget the concept of the object in the memory. Using `null` instead of `{}` can solved this issue. 

- Thank for writing a shorter code , what makes my code longer is : I tried to deal with a situation like this: 
`{"todoes:{"description":"task 1"}` so if the key is something else that `todo`, and my way makes things more complicated. 

- object equality: is something I really  need to review . Thanks

week3/homework/src/utilities.js
Adham-Elias-Botours: I have to be very careful about that point.

week3/homework/src/utilities.js
Adham-Elias-Botours: that makes sense: the constructor has to construct it with all what it needs. so also with ID.

week3/homework/src/utilities.js
Adham-Elias-Botours: 2 good ideas . 
So in general it is good to decrease throwing errors while another options are there.

week3/homework/src/utilities.js
Adham-Elias-Botours: `deleteToDo`: to delete singel ToDo item by its ID.

`deleteToDos`: to delete/clear the list , delete all the ToDos
Did I understand your question?

week3/homework/src/app.js
Adham-Elias-Botours: At that moment, I ran out of names. 
You understood my files/folders structure and your suggestion is perfect with it.

week3/homework/src/utilities.js
Adham-Elias-Botours: ops, in this case I am sending it as text without mentioning that. Now I understand why I get that text style when the ToDos list appears in the PostMan

Adham-Elias-Botours:
@Meno-101 Thanks for your time on  reading and reviewing my code.

PR #481, opened 2019-10-09
Node.js Week1 Homework

ihk2018:


week1/homework/src/server.js
```diff
@@ -5,11 +5,37 @@ const http = require('http');
 /* `createServer` MUST return an instance of `http.Server` otherwise the tests
  * will fail.
  */
+
 function createServer(port) {
   let state = 10;
 
+  function getResponse(response, state) {
```
shrynx: the name of this function is confusing
this is not getting any response
rather it is constructing/making a response

ihk2018: dear @shrynx thank you for the feedback, based on that i changed the name of this function.

PR #483, opened 2019-10-10
Week1-HomeWork

husenjanobul:


week1/homework/src/server.js
```diff
@@ -10,6 +10,21 @@ function createServer(port) {
 
   const server = http.createServer((request, response) => {
     // TODO: Write your homework code here
+    response.setHeader('Content-Type', 'application/json');
+    if (request.url === '/' || request.url === '/reset') {
```
shrynx: this looks bit buggy
from what i understand
`'/'` should return the current state and not reset it

husenjanobul: Thank you for feedback . i fixed it .

PR #485, opened 2019-10-11
Mehmet Ekinci-Week1

mehmetekinci:


week1/homework/src/server.js
```diff
@@ -9,7 +9,38 @@ function createServer(port) {
...
+        responseJsonObject.state = state;
+        response.write(JSON.stringify(responseJsonObject));
+        break;
+      }
+      case '/add': {
+        responseJsonObject.state = ++state;
+        response.write(JSON.stringify(responseJsonObject));
+        break;
+      }
+      case '/subtract': {
```
limoragni: This does not break the functionality. But should be `substract` and not `subtract` just a small type :)

mehmetekinci:
hi @limoragni. I made it a bit clear but when I make substract, tests did not passed. also subtract  is on readme.md file which is in week1 folder. I left it subtract. thank you again.

PR #486, opened 2019-10-11
nodejs week1

malnaqeeb:


week1/homework/src/server.js
```diff
@@ -5,13 +5,33 @@ const http = require('http');
 /* `createServer` MUST return an instance of `http.Server` otherwise the tests
  * will fail.
  */
+function  getResult(status, response, result){
```
shrynx: the name getResult is a bit odd
Since this is not getting the result from somewhere
rather it is making/building response

malnaqeeb:
Thanks @shrynx for your review, I have implemented the mentioned feedback

PR #489, opened 2019-10-16
Node.js Week1 Homework

FarukTercan:
Hi, I couldn't upload week1 homework on time due to my health problem.
 I am glad if you could review it.

### DIFF
--- a/week1/homework/src/server.js
+++ b/week1/homework/src/server.js
@@ -2,14 +2,35 @@
 
 const http = require('http');
 
-/* `createServer` MUST return an instance of `http.Server` otherwise the tests
- * will fail.
- */
 function createServer(port) {
   let state = 10;
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    let serverResponseObject = {};
+    response.setHeader('Content-Type', 'application/json');
+
+    if (request.url === '/state') {
+      serverResponseObject.state = state;
+    }
+ else if (request.url === '/add') {
+      state++;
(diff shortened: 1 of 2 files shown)


week1/homework/src/server.js
```diff
@@ -2,14 +2,35 @@
...
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    let serverResponseObject = {};
+    response.setHeader('Content-Type', 'application/json');
+
+    if (request.url === '/state') {
+      serverResponseObject.state = state;
+    }
+ else if (request.url === '/add') {
```
limoragni: The indentation of the `else if` clauses are wrong. They should be aligned with the first `if`

week1/homework/src/server.js
```diff
@@ -2,14 +2,35 @@
...
 
-/* `createServer` MUST return an instance of `http.Server` otherwise the tests
- * will fail.
- */
 function createServer(port) {
   let state = 10;
 
   const server = http.createServer((request, response) => {
-    // TODO: Write your homework code here
+    let serverResponseObject = {};
```
limoragni: You are never re-assigning `serverResponseObject` to a different value. It always continues being the same object. You are then changing the properties of the object but not changing the variable. For this reason it's better to use `const` instead of 'let'

week1/homework/src/server.js
FarukTercan: Thanks, I fixed it by ignoring prettier.

PR #490, opened 2019-10-16
week1

Ann-Grabetski:
Hi! 
My code does not pass the test. But I have no idea why!

### DIFF
--- a/week1/homework/src/server.js
+++ b/week1/homework/src/server.js
@@ -1,20 +1,33 @@
 'use strict';
 
-const http = require('http');
+let state = 10;
 
-/* `createServer` MUST return an instance of `http.Server` otherwise the tests
- * will fail.
- */
-function createServer(port) {
-  let state = 10;
+function handleRequest(request, response) {
+  try {
+    switch (request.url) {
+      case '/add':
+        state++;
+        break;
+      case '/subtract':
+        state--;
+        break;
+      case '/reset':
+        state = 10;
+        break;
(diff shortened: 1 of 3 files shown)


Ann-Grabetski:
I turned back with server code and it helped. But still, one test doesn't pass. I don't understand why.

week1/homework/src/server.js
```diff
@@ -9,7 +9,30 @@ function createServer(port) {
...
+        case '/reset':
+          state = 10;
+          break;
+      }
+      const stateJSON = { 'state': state };
+      response.setHeader('Content-Type', 'application/json');
+      response.write(JSON.stringify(stateJSON, null, 2));
+    }
+    catch (err) {
+      response.statusCode = 404;
```
limoragni: The code looks good! There is one mistake that's preventing the test to pass. You have a `try/catch` where you are handling errors. In case there is an error you are then returning a `404`. The problem is that the `404` needs to be returned, not when there is an error, but when there are no matches on the URL. 

What's happening now if you go to `http://localhost:3000/blahblah` or any URL that doesn't match is that you are getting the state. Because there are no matches in the `switch` but no errors, so it goes to sending the `state` back. 

I like that you are trying to catch errors. What you need now is a `default` in your `switch` that sends the `404` and then change what you send in the catch. You can send a `500` saying there was an unexpected error. That's not part of the homework though, but since you had the code there already I think is .a nice addition.

PR #491, opened 2019-10-17
Add node.js week2 homework

EsraZengn:
Here is my week 2 homework.

### DIFF
--- a/week2/homework/src/index.js
+++ b/week2/homework/src/index.js
@@ -1,3 +1,114 @@
 'use strict';
+{
+  const { readList } = require('./readFile.js');
+  const { writeList } = require('./writeFile');
+  const program = require('commander');
 
-// TODO: Write the homework code in this file
+  const todoListPath = './todoList.json';
+
+  function displayHelp() {
+    program
+      .version('0.1.0')
+      .command('node <file-name> ', 'output usage information')
+      .command('list <file-name>', 'list all the items')
+      .command('add <file-name> [item]', 'add item to the list')
+      .command('update <file-name> [index] [new-item]', 'update list')
+      .command('remove <file-name> [index]', 'remove specified item')
+      .command('reset <file-name>', 'reset the list')
+      .parse(process.argv);
+  }
+
+  const initializeTodoList = () => {
+    const todoList = [];
+    writeList(todoListPath, todoList);
+    return todoList;
+  };
+
+  const addTodoListItem = async newTodo => {
+    try {
+      let { todoList } = await readList(todoListPath);
+      if (todoList === undefined) {
+        todoList = initializeTodoList();
+      }
+
+      todoList.forEach((todo, index) => {
+        todo.id = index + 1;
+      });
+      let index = todoList.length + 1;
+      todoList.push({ id: index, todo: newTodo });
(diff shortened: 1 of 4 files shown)


week2/homework/src/index.js
```diff
@@ -1,3 +1,114 @@
...
+  };
+
+  const addTodoListItem = async newTodo => {
+    try {
+      let { todoList } = await readList(todoListPath);
+      if (todoList === undefined) {
+        todoList = initializeTodoList();
+      }
+
+      todoList.forEach((todo, index) => {
```
limoragni: I see this `forEach` is re-assigning all ids to the value they already have. What do you think is the need for this loop?

week2/homework/src/index.js
```diff
@@ -1,3 +1,114 @@
...
+        todo.id = index + 1;
+      });
+      let index = todoList.length + 1;
+      todoList.push({ id: index, todo: newTodo });
+      writeList(todoListPath, todoList);
```
limoragni: Would be nice to receive a message when a todo is added successfully ;)

week2/homework/src/index.js
```diff
@@ -1,3 +1,114 @@
...
+  };
+
+  const removeTodoListItem = async itemToRemove => {
+    let { todoList } = await readList(todoListPath);
+    try {
+      if (!todoList || todoList.length === 0) {
+        console.log('The list is empty!');
+      } else if (itemToRemove <= todoList.length && itemToRemove > 0) {
+        todoList = todoList.filter(todo => todo.id !== itemToRemove);
+        todoList.forEach((todo, index) => {
```
limoragni: I see, this makes more sense here. When you delete a `todo` you want them to continue having consecutive ids. That's an interesting approach and totally valid. 

Keep in mind that normally when you have internal ids on a database (in this case the json file is the DB) it's very uncommon to change the ids. Let's say I first check the list and think, ok I want to delete Ids 3, 6, 8. Then I go and run the delete commando three times with each of those numbers, I would be deleting the wrong todos, does that make sense?

In case you also want to have a number for the positions of the todos in the list you could show to the user a `position` property and show `index + 1`.

In this case I would say you are ok removing the `forEach`. Let me know if that sounds ok, or you have any doubts.

week2/homework/src/index.js
```diff
@@ -1,3 +1,114 @@
...
+    writeList(todoListPath, todoList);
+  };
+
+  const [_, __, ...commandItems] = process.argv;
+
+  if (commandItems.length === 0 || commandItems[0] === 'help') {
+    displayHelp();
+  } else if (commandItems[0] === 'list') {
+    readList(todoListPath).then(({ todoList }) => {
+      if (todoList && todoList === []) {
```
limoragni: Javascript doesn't like this `todoList === []` even `[] === []` is `false`. I know, it makes little sense, but these are some of the quirks of the language. 

Because of that is that the `list` command is not working. 

I guess you maybe where expecting that to be checking whether this is an array or not? 
If you want to check a variable to know if it's an array you can use `Array.isArray(todoList)`
In this case it seems you want to check if the array is empty, you can use the `length` property for that.

PR #502, opened 2019-10-23
finished homework for node week03

Laurelinduilwen:
This is Arya from class-22, this was a missing homework of mine that I was finally able to submit :(

tried to preserve the structure of the lecture app as much as possible

### DIFF
--- a/week3/homework/src/actions/clearTodos.js
+++ b/week3/homework/src/actions/clearTodos.js
@@ -1,0 +1,12 @@
+'use strict';
+
+async function clearTodos(todo, req, res) {
+  try {
+    await todo._save([]);
+    res.status(201).send({ Succeeded: 'Obliterated all todo items >=3' });
+  } catch (error) {
+    res.status(404).send({ error: '404 Not found :(' });
+  }
+}
+
+module.exports = clearTodos;
--- a/week3/homework/src/actions/markAsDone.js
+++ b/week3/homework/src/actions/markAsDone.js
@@ -1,0 +1,23 @@
+async function markAsDone(todo, req, res) {
+  try {
+    const todoList = await todo.readList().then(todoList => JSON.parse(todoList));
+    const index = parseInt(req.params.id);
+    if (index > 0 && index <= todoList.length) {
+      if (todoList[index - 1].done === false) {
+        todoList[index - 1].done = true;
+        await todo._save(todoList);
+        res.status(404).send({ Success: 'Todo has been marked as DONE' });
+      } else if (todoList[index - 1].done === true) {
+        todoList[index - 1].done = false;
+        await todo._save(todoList);
+        res.status(201).send({ Success: 'Todo has been marked as UN-DONE' });
+      }
+    } else {
+      res.status(400).send({ Failed: 'Please insert a valid id' });
+    }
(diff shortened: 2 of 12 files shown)


week3/homework/src/actions/clearTodos.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+async function clearTodos(todo, req, res) {
+  try {
+    await todo._save([]);
+    res.status(201).send({ Succeeded: 'Obliterated all todo items >=3' });
```
gajduk: 🤣🤣

week3/homework/src/actions/markAsDone.js
```diff
@@ -0,0 +1,23 @@
+async function markAsDone(todo, req, res) {
+  try {
+    const todoList = await todo.readList().then(todoList => JSON.parse(todoList));
+    const index = parseInt(req.params.id);
+    if (index > 0 && index <= todoList.length) {
```
gajduk: 👍  good job on catching Index out of bounds. Here is a small tip for the future. To avoid nesting if statements like you have done, you can use a so called *early return*
```
if ( index < 0 || index >= todoList.length ) {
   res.status(404).send({ Failed: 'Please insert a valid id'});
   return;
}
// notice how this is no longer nested
if ( todoList ...
```

week3/homework/src/actions/readTodo.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+async function readTodo(todo, req, res) {
+  try {
+    const todoList = await todo.readList().then(todoList => JSON.parse(todoList));
+    const index = parseInt(req.params.id);
+    if (index > 0 && index <= todoList.length) {
+      res.status(200).send(todoList[index - 1]);
+    } else {
+      res.status(400).send({ Failed: 'Please provide a valid id ' });
```
gajduk: this seems to be repeated also above. This means it needs to be put in a separate function. Can you think of a way to achieve this?

week3/homework/src/todo.js
```diff
@@ -0,0 +1,48 @@
...
+        if (err) {
+          if (err.code === 'ENOENT') console.log('file not found');
+          return reject(err);
+        }
+        return resolve(todoList);
+      }),
+    );
+  }
+
+  async delete_(id) {
```
gajduk: is this used anywhere?

week3/homework/src/index.js
```diff
@@ -1,3 +1,29 @@
...
+const TODO_SLUG = 'todos';
+
+//keeping this here from the lecture for convenience :3
+app.get(`/${TODO_SLUG}`, readTodos.bind(null, todo));
+//reads a single todo item
+app.get(`/${TODO_SLUG}/:id`, readTodo.bind(null, todo));
+//DESTROYS EVERYTHING :(
+app.delete(`/${TODO_SLUG}`, clearTodos.bind(null, todo));
+//implementing this as a toggle since the assignment also wants the code to be DRY
+app.put(`/${TODO_SLUG}/:id/toggle-done`, markAsDone.bind(null, todo));
```
gajduk: This URL is not REST-like. Can you guess why?

PR #503, opened 2019-10-25
Msonmez homework week3

msnmz:


week3/homework/src/actions/deserializeTodo.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+async function deserializeTodo(request) {
```
gajduk: 👍 good job on validation. Since this function is only validating perhaps it needs a different name 😄

week3/homework/src/actions/deserializeTodo.js
```diff
@@ -0,0 +1,15 @@
...
+  const { todo } = request.body;
+  if (todo == null) throw new Error('todo not set');
+
+  if (todo.description != null) todo.description = todo.description.trim();
+
+  if (todo.description == null || todo.description.length === 0)
+    throw new Error('description not set');
```
gajduk: what would happen if a user tries to create a new todo with an empty description?
Is this the correct behaviour?

week3/homework/src/todo.js
```diff
@@ -0,0 +1,95 @@
...
+
+    await this._save(todos);
+
+    return todo;
+  }
+
+  read() {
+    return new Promise(resolve => {
+      fs.readFile(this._filename, DEFAULT_ENCODING, (error, data) => {
+        if (error) return resolve([]);
```
gajduk: it turns out that there are other reasons why reading a file can fail, it is not just when the file is not there.
It would be best to explicitly check if the file exists instead of abusing the error like this.

week3/homework/src/todo.js
```diff
@@ -0,0 +1,95 @@
...
+
+  async mark(id, done) {
+    const todos = await this.read();
+
+    const todo = todos.find(t => t.id === id);
+    if (!todo) {
+      const error = new Error(`To-do with ID ${id} does not exist`);
+      error.code = 'not-found';
+      throw error;
+    }
```
gajduk: this code is repeated just 10 lines above 😢

PR #556, opened 2021-02-02
Homework week 1 - Mohammed Saif

MoSaif00:
In this week homework, I have worked on three exercises : 

1. Pad-left using exports from padleft.js file.
2. Pad-left using the pad-left module in the npm packages.
3. created my first webserver.

And for the week Project, was about HackYourTemperature and I started the project by installing some npm modules like express, Axios ...etc and created an express server


MoSaif00:
I made pull request by mistake here

PR #563, opened 2021-09-21
Update MAKEME.md

ObadaElSharbatly:


robvk:
Thanks for offering up a suggestion, but figuring out that that is the way to write that code is part of the assignment. With the idea being that this is how you would describe that line of code. Students will need to learn that the words 'method' and 'function' are used interchangeably and that 'pass in the' means to use it as a parameter.

If we just give students the line of code, it will be a copy/paste and nothing will be learned.

PR #637, opened 2024-05-09
feat(week3): Introduce a database helper

sarlam:
Summary
=======

We noticed through the course of the previous run that students got overwhelmed by having to implement both the auth layer as well as dealing with persisting them.

Since the next module is about database, this PR is a proposal to introduce a `database` helper so they don't have to deal with the database right now.

## What's in the box

From a student perspective, it adds a bit of reading in the README to understand how to use the database.

It also reshapes the readme to have better sections to help navigate it.

From our standpoint, it introduces lokijs and uuid in the package.json and a database.js file that allow both to create a persistent or in memory database.

### DIFF
--- a/week3/prep-exercise/README.md
+++ b/week3/prep-exercise/README.md
@@ -1,12 +1,23 @@
-# Prep exercise
+# Server Prep exercise week 3
 
-## Server
+## Goals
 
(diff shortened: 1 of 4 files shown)


stasel:
Looks great! Thanks for your contribution

PR #679, opened 2025-11-05
Ivan w2 node js assignment

Sky-A-Fox:
Tasks finished.
Test passed
I tryed to hide my api key.
At my PC everything works. but I don't know if will works during check


assignments/hackyourtemperature_W2/sources/keys.js
```diff
@@ -0,0 +1,9 @@
+// sources/keys.js
+import dotenv from "dotenv";
```
durw4rd, line 2: Great that you are using dotenv for loading the env variables in your app!

assignments/hackyourtemperature_W2/sources/keys.js
```diff
@@ -0,0 +1,9 @@
+// sources/keys.js
+import dotenv from "dotenv";
+dotenv.config(); // Загружает переменные из .env === loads variables from .env
+
+const keys = {
+  API_KEY: process.env.API_KEY || "" // fallback на пустую строку === fallback to empty string
```
durw4rd, line 6: An empty string as a fallback value for API credentials can be confusing to people not familiar with the app. Consider throwing an error or log warning to make it more explicit what's happening.

assignments/hackyourtemperature_W2/server.js
```diff
@@ -0,0 +1,4 @@
+import app from "./app.js";
+// Start the server === don't add funktionality in this file
+const PORT = 3000;
```
durw4rd, line 3: Nit: Port is also sometimes stored as an environment variable - handy as it gives more flexibility for the deployment.

assignments/hackyourtemperature_W2/app.js
```diff
@@ -0,0 +1,43 @@
...
+app.use(express.json());
+
+// Проверка работы сервера === test route
+app.get("/", (req, res) => {
+  res.send("hello from backend to frontend!");
+});
+
+// Получение погоды по названию города ===  get weather by city name
+app.post("/weather", async (req, res) => {
+  const cityName = req.body.cityName;
```
durw4rd, line 14: Consider validation for empty string or non-string type input.

assignments/hackyourtemperature_W2/app.js
```diff
@@ -0,0 +1,43 @@
...
+    // Запрос к OpenWeather API === request to OpenWeather API
+    const response = await fetch(
+      `https://api.openweathermap.org/data/2.5/weather?q=${cityName}&appid=${keys.API_KEY}&units=metric`
+    );
+    const data = await response.json();
+
+    // Если город не найден === if city is not found
+    if (data.cod !== 200) {
+      return res.status(404).json({ weatherText: "City is not found!" });
+    }
```
durw4rd, line 31: This assumes that any non-200 `cod` value has to mean a 404, city not found error. It may as well be the case (I don't know this), but in case other reasons can lead to a non-200 response code, a more robust way of handling these responses would be passing the `data.cod` value to the return statement to provide help with debugging.

assignments/hackyourtemperature_W2/app.js
```diff
@@ -0,0 +1,43 @@
...
+
+    // Отправка температуры пользователю === send temperature to user
+    const temperature = data.main.temp;
```
durw4rd, line 34: Consider: Use optional chaining to prevent crashes if the API response structure is unexpected
 `data?.main?.temp` instead of `data.main.temp`

durw4rd (approved):

durw4rd (approved):

durw4rd:
Good job with the assignment, Ivan. It meets the requirements; I've added a few comments,s but none of them are critical. 
What is critical, though, you've made the PR against the wrong repository. Can you please make the PR for week2 against https://github.com/HackYourAssignment/Node.js-Cohort54? I'll accept it there once ready.

